PR #4, opened 2019-10-22
Ensure list items used as nav-items do not have a bullet points.

bernhard-hofmann:
Ensure list items used as nav-items do not have a bullet point. Fixes bullet point shown beside the `Sign In` button on smaller screens.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -285,6 +285,10 @@ nav .navbar-nav li a.nuxt-link-active[data-v-314f53c6] {
   background-color: #61ae24 !important;
 }
 
+nav li.nav-item {
+  list-style-type: none;
+}
+
 nav .navbar-nav li a {
   color: #cdcdcd !important;
 }


edwh:
The fix works - thanks - but I think it would be better implemented using existing BootstrapVue classes.  Probably the ul/li which contains this button should be a b-navbar-nav/b-nav-item; we shouldn't really be using raw uls.  Could you adapt for that?

bernhard-hofmann:
I didn't make the `ul` or `li` and assumed whoever did intended it to be that way.

Should the contributor guidelines include seeking direction on changes before submitting a PR to avoid these misunderstandings?

edwh:
I did the original, but it was probably copied from the older code and I 
failed to change it.

Not sure what the contributor guidelines would say?  "Don't assume that 
something that looks wrong is there for a good reason unless there's a 
comment"?

edwh:
I've done a checkin which should fix this.  Thanks for spotting it.

PR #6, opened 2019-10-23
Remove some element selectors as they aren't good practice

JayyajGH:
There are quite a few element selectors in the code which is bad practice.  This is a first attempt at removing a few and using bootstrap selectors in their place.  The colour ones will have to wait until there is a standardised colour system in place.


edwh:
Thank you, that's definitely better.

PR #9, opened 2019-10-25
Start to consolidate some of the colors into a single location

JayyajGH:
What are your thoughts on this style?  Ideally it would be good to use CSS custom properties for this but that would involve a polyfill for older browsers and probably a bit more thought.  I've used this method elsewhere and it seems to work well.  Because they are Sass variables they don't create any extra code.
Basically my plan would be to have all colours defined in a single location.  Then you will probably see the massive variety of similar colours and we can then consolidate them in a common set.  If you're happy with this then let me know and I'll continue with the rest before you merge this PR.

### DIFF
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -16,6 +16,7 @@ $color-green--dark: #006b50;
 $color-green--dark-alt: #006400;
 $color-green--dark-alt2: #3b8070;
 $color-green--dark-alt3: #008000;
+$color-green--dark-alt4: #00523d;
 
 $color-gray--faded: #6c757d70;
 $color-gray--x-light: #f5f5f5;
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -39,3 +39,9 @@ $color-red: #FF0000;
 
 $color-white: #FFF;
 $color-black: #000;
+
+$color-white-opacity-50: rgba(255, 255, 255, 0.5);
+$color-white-opacity-75: rgba(255, 255, 255, 0.75);
+$color-black-opacity-33: rgba(0, 0, 0, 0.33);
+$color-black-opacity-59: rgba(0, 0, 0, 0.59);
+$color-black-opacity-60: rgba(0, 0, 0, 0.6);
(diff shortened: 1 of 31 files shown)


edwh:
Thanks.  I had a look and the approach looks fine - I'm in favour of 
consolidating them into a single place for the precise reasons you give.

The only change I would suggest you consider is naming by purpose not by 
colour.  For example, the green "success" type colour is supposed to be 
the primary call to action, white is supposed to be something you only 
see if you look for it, and the others in between are probably a bit of 
a muddle.
By using names which are about what we're trying to achieve with the 
colour, we'd probably a) think more clearly about that when we use them 
and b) be able to change the actual colours without then having to 
remember that dark green is now purple, or whatever.

JayyajGH:
I used to name colours like that and then I changed to this kind of style.  Lots of the influential coder/designers like Chris Coyier switched to this style a few years back.  I personally much prefer using this method now.

https://css-tricks.com/what-do-you-name-color-variables/

I think that if your component architecture is set up correctly then you wouldn't really need to think about that stuff very often.  e.g. for a button I usually have something like button--primary, button--secondary and button--tertiary.  I never have to think about the colours as once the button component is created then I forget about it. That goes for the majority of my HTML.

The argument against your (b) is that if your colours are defined in a single location then it takes 10 seconds to do a global find and replace to change dark-green to purple.

How about this....
At the moment I think you probably have about 40-50 different colours.  There is no way that we can name them anything useful.  How about phase 1 I continue down the route I'm going.  This is likely to descend into grey-2, grey-3, grey-4.  Then phase 2 is consolidate them down to a manageable set.  And phase 3 if you still prefer it is naming then by purpose.

Sorry - I should have warned you I was quite opinionated about HTML and CSS. :-).  Don't get me started on accessibility....  I do understand that it is your project though and more than happy to go with what you decide.

edwh:
Well argued.  I defer to your experience - stick with the colour names.

Opinionated is good.  I want that.

Shall I merge this one, then?

JayyajGH:
OK this is the first stab.  Next I'll try and sweep up and RGBs and any inline style definitions too.  Then on to trying to reduce the number of colours as so many of them are so similar.  Maybe take a look at this and if you're happy merge this and I'll create another PR for any other work.

edwh:
Will review but probably not till tonight.

PR #11, opened 2019-10-28
Move some classes to a more suitable location

JayyajGH:
What I`m attempting to do is basically empty out as many global classes into their local components as I can. This might lead to some more components getting created where appropriate. This will probably be done a couple of files at a time to make sure nothing gets broken.  There appears to be quite a few specificity issues so it might be tricky.  !important is used a lot.

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -11,10 +11,6 @@ $bootstrap-lg: 1200px;
   color: $colour-success-fg !important;
 }
 
-.border-success {
-  border-color: $colour-success-bg !important;
-}
-
 .card-header.bg-success {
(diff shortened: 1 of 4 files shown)


edwh:
!important is probably overused. There are some times when I definitely need it, but I also tend to slap it in without thinking about it just in case.

JayyajGH:
I`ve always tended to work on greenfield sites without using any of the heavyweight CSS frameworks and so I`ve always got away with not using it - unless I am writing utility first CSS.  Important is definitely required sometimes but that uses just points towards an issue somewhere else.  Which might be external and you not have any control over.  I`ve not used bootstrap before so I`m sure I might hit the same issues soon.

PR #12, opened 2019-10-29
Enhancements/fix scoped class  issue

JayyajGH:
Using the deep selector means that the class is preserved when used in a child component.

See here for a description:

[(https://vue-loader.vuejs.org/guide/scoped-css.html#mixing-local-and-global-styles)]


JayyajGH:
This wasn't the correct branch....

PR #14, opened 2019-10-30
Add a deep selector so that the class can be used in a child component

JayyajGH:
Use a deep selector so that the class can be used in a child component and not removed by the postcss process.

See here for a description
[https://vue-loader.vuejs.org/guide/scoped-css.html#child-component-root-elements](url)


edwh:
Nice.

PR #15, opened 2019-10-30
Move iteminp class to where it is used

JayyajGH:
This class is only used with the autocomplete component so it should live with it.  There is then no issues with postcss removing it either.


JayyajGH:
Both postcodelist and iteminp classes now live on the component they are used on.

PR #16, opened 2019-10-30
Move the autocomplete classes into the autocomplete component

JayyajGH:
Move all the classes from the global autocomplete.css file into the component itself.

Was there any reason I`m missing for having these in a separate file?  I`ve moved them and the consumers of the autocomplete component still appear to function correctly.


edwh:
It originally derived from a third party component which I adopted 
because I wanted to change it - there's a credit in Autocomplete.vue 
around line 57.  The original component had a separate CSS file, I 
think, but it's better to have them in there.

PR #21, opened 2019-10-31
Line the postcode elements up correctly

JayyajGH:


JayyajGH:
Sorry, forgot a description...

So i've used flexbox just to line the two elements up.  I've tried this in a few different size configurations and it works ok.   You might want to give it a bit of a test in case I've missed a use case.  I've also given the input element a bit of breathing space so there is now room for the focus indicator.

edwh:
Nice to see TODOs disappearing...

PR #25, opened 2019-11-01
Create sidebar components

JayyajGH:
The sidebars were duplicated in three different files so I`ve put them into components.  We could go further and create a common sidebar wrapper in the future...

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -478,10 +478,6 @@ legend {
   padding-right: 0px;
 }
 
-.sidebar {
-  height: calc(100vh - 100px);
-}
-
 .mx-datepicker-popup {
   z-index: 2000 !important;;
 }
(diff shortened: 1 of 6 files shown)


edwh:
I can see the reasoning for this.  However there are some more subtle 
cases.  For example on the Volunteer Opportunity page, I have been 
thinking we will probably still want to show the job ads in the right 
sidebar, but not the volunteer ops because that's in the central part.  
Similarly for Community Events pages.

So I think you probably need some props which allow you to turn on/off 
which components within the sidebar appear.

JayyajGH:
OK cool.  I`ll add some in.

JayyajGH:
Do you want all sidebar sections to be controlled by props? i.e. currently all four?

edwh:
Yeah, while you're in there.  Probably overkill for the left bottom box 
but might as well.

Thanks.

JayyajGH:
I`ve just done it for the right hand sidebar at the moment.  If we do it for the left then the menu section floats to the top which presents a new set of problems.

edwh:
Ok, that's fine for now.

PR #26, opened 2019-11-01
Change the alert box into a notification section

JayyajGH:
Change the alert box into a regular div.

For now I`ve just created something locally in one place just to check everything works as expected.  The next stage would be to analyse what other alert boxes should be changed to see what kind of component we could create.

NB: See how using flexbox makes the HTML much simpler and easy to read.  It just sorts out all of the rows, columns, positioning etc. for you.  It`s also halved the number of elements required.

### DIFF
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -20,6 +20,8 @@ $color-gray--light: #CDCDCD;
 $color-gray--base: #A9A9A9;
 $color-gray--dark: #808080;
 
+$color-blue--x-light: #D1ECF1;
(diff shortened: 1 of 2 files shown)


edwh:
Yes, the flex stuff is much neater.  I will endeavour to mend my obsolete ways.

I definitely think we should have a separate component, though, but that can be later.  Will merge.

JayyajGH:
Yeah it definitely sounds like a component.  I think it just needs some thought into what props etc. it needs.  Different colours? Options? etc.  I`ve added it to my list...

PR #27, opened 2019-11-01
Refactor the wizards into separate components.

JayyajGH:
Move the duplicated wizard elements into separate components.  These were all in 2-3 places each. This should make the existing pages which are quite large, a little bit easier to read.  Next step would be to reduce this down to a single wizard component and then all the global CSS classes relating to this component can be moved into it. #slowlykillingglobalcss


edwh:
Good.

I notice you're using kebab case for the filenames (though not the use 
in the template).  The rest of the component names are PascalCase.  I 
think both are ok by Vue 
(https://vuejs.org/v2/guide/components-registration.html#Component-Names) 
- am I missing something?

JayyajGH:
Take a look at the Vue styleguide I sent you.  I`m pretty sure that this is the way that they recommend.

JayyajGH:
But happy to update them if that is what you prefer...  Especially if both are valid and all the others are like that.

JayyajGH:
Oh yeah, just seen your link.  Doh

edwh:
Yeah, I think both are legit, so if you could stick with Pascal, that 
would be good.

JayyajGH:
OK so I`ve now binned the three separate components and merged them into a single one with props.  I`ve also named it as per the project standard.

PR #29, opened 2019-11-06
Sort out the date layout so it works across all breakpoints

JayyajGH:
Fix the date layout in all breakpoints.

This resolves the following problems:

- I`ve moved the labels above the fields as this gives them more space as previously they would break almost instantly on smaller screens. Having the label about the field helps people scan better.
- Was previously set up as a bootstrap card which it isn`t really
- Remove the placeholders - Generally placeholders aren`t great for various reasons like accessibility, people can more easily scan for empty fields rather than fields already containing stuff, auto translate tools won`t translate them, when you start typing you lose the text etc.  The other fields should probably move the placeholder text into smaller text above or below the field

### DIFF
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -9,6 +9,7 @@ $colour-header: #1d6607;
 $colour-primary: #00A1CB;
 
 $color-green--light-alt: #00FF00;
+$color-green--lighter: #CDE4DA;
 $color-green--light: #E6ffE6;
 $color-green--medium: #84CF96;
 $color-green--dark: #3B8070;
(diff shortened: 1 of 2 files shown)


edwh:
Interesting about placeholder and accessibility - didn't realise that.  Thanks.

JayyajGH:
My main annoyance about placeholders ( especially when used instead of labels ) is that as soon as you start typing the text is lost.

This is a great article on the subject from the NN group:
https://www.nngroup.com/articles/form-design-placeholders/

edwh:
Yes, I (now) agree with you.  I'd thought of them as a handy way to save 
screen real estate, but I hadn't thought properly about the 
visual/cognitive aspect.  Very useful, thanks.

PR #30, opened 2019-11-06
Update to the latest version of the FAQ package which doesn't centre

JayyajGH:
My PR on the FAQ package has now been accepted and so the container no longer centres by default.  Therefore we can now bin the overriding of the internal class.


edwh:
Unexpectedly.  Good.

PR #31, opened 2019-11-06
Enhancements/simplify delete button

JayyajGH:
Simplify the delete button and make it tab-able.

Because there is now room, I`ve left the text in at all breakpoints.  Generally this is considered a good thing from a UX perspective.  If you don`t like the look then I can remove it on larger screens but then we would need to do some aria-fu.

I`ve had to remove the "important" for the white button border.  I`ve tested this hasn`t caused issues wherever I can but I might have missed something that we need to fix it later.

References:
https://css-tricks.com/do-you-need-an-icon-only-button-without-screwing-up-the-accessibility/
https://www.sarasoueidan.com/blog/accessible-icon-buttons/

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -78,7 +78,7 @@ $bootstrap-lg: 1200px;
   color: $color-black !important;
   text-shadow: none !important;
   -webkit-font-smoothing: antialiased;
-  border: 1px $color-black solid !important;
+  border: 1px $color-black solid;
 }
 
 .button--green {
--- a/components/StartEndDate.vue
+++ b/components/StartEndDate.vue
@@ -38,7 +38,7 @@
     <div>
       <!-- TODO Hide button if only one date present -->
       <b-btn variant="white" class="delete__button" size="sm" @click="$emit('remove', index)">
-        <v-icon name="trash-alt" class="delete__icon" title="Delete this date" aria-hidden="true" />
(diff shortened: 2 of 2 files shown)


edwh:
Thanks.  I have (as you will have noticed) given zero thought to tabbing.

JayyajGH:
Ha! Everyone works differently I guess.  I prefer to sort out tabbing, accessibility, breakpoints etc. all at the same time rather than at the end.  I`ve always worked using Agile and so it`s either complete or it`s not complete.

edwh:
I work on Freegle in a weird way.

Because we are chronically short on resource, I do a 90/10 rule thing 
and cut corners.  The last few per cent of what I'd normally do in a 
normal working environment to do it right takes sufficiently long that I 
can get better bang for buck on something else which desperately needs 
love.

Plus I'm trying to get this new version out to the point where I can 
give it to volunteers as a beta, because they'll then spend an elapsed 
few weeks playing with it.  During that time I can fix up stuff they'll 
probably never look at.

Alternatively, I'm a workshy fop who produces overintellectualised 
excuses.

But I'm very happy for you to fix up some of that stuff as you find it.

JayyajGH:
Ha!  Yeah I completely get it.  A half working product is better than no product.  It`s amazing what you`ve managed to  create with just one person in a short time.

PR #32, opened 2019-11-06
Fix the issue with deleting dates

JayyajGH:
I tested this before I refactored the startenddate component and the issue was still there.  It looks like the array is correct but the display is doing something weird and not updating correctly.

### DIFF
--- a/components/StartEndDate.vue
+++ b/components/StartEndDate.vue
@@ -37,6 +37,7 @@
     </div>
     <div>
       <!-- TODO Hide button if only one date present -->
+      <!-- TODO Delete button isn't working correctly - Add three dates and delete the first one to recreate -->
       <b-btn variant="white" size="sm" @click="$emit('remove', index)">
         <v-icon name="trash-alt" title="Delete this date" aria-hidden="true" />
         <span class="delete__label">Remove</span>
--- a/components/StartEndDate.vue
+++ b/components/StartEndDate.vue
@@ -68,6 +68,14 @@ export default {
       endd: null
     }
   },
+  watch: {
+    start: function() {
+      this.startd = this.start
+    },
+    end: function() {
+      this.endd = this.end
+    }
+  },
   mounted: function() {
     this.startd = this.start
     this.endd = this.end


edwh:
Maybe try replacing the splice with:

`this.$delete(this.editableDates, index)`

JayyajGH:
The splice is working ok.  The data in the component is ok after the delete too.  Looks like the date-picker component isn`t updating it`s display.

JayyajGH:
Hmmm.  What is the startd stuff all about?  Looks like that isn`t being updated.  The start prop is updated but the startd data isn`t

edwh:
That's usually a reactivity issue, which can be caused by how the data 
is set.

Anyway, leave the TODO on there and I'll have a look at some point.

edwh:
I think that's to avoid mutating the props.  Perhaps the change() method 
needs to set startd/endd.

JayyajGH:
The delete button removes the date from the editable dates collection. The display fields (date-picker) get there data from the startd and endd data values which only seem to be updated on mount.

JayyajGH:
Startd and endd are already set by the component.  The start and end props are used to set up startd and endd on mount.  I`m thinking the only way to make the internal data values change is to set them from a computed property or a watch.  I`ll have a think...

edwh:
I'll have a look when I finish the chunk I'm doing.

JayyajGH:
I hate being stumped by stuff...  Thinking this through you need to update the local data value when the prop changes.  The only way I can think of doing that is with a watcher.

e.g.
`
watch: {
   start: function() {
      this.startd = this.start
    }
}
`

This seems to work correctly.  I`ll update the  PR for you to have a look at when you`re next free...

edwh:
But what is it that is changing the values?

JayyajGH:
If you click delete then it raises an event to the parent component - StartEndCollection - that deletes the entry from the array (with splice).  The prop is then updated in the StartEndDate component but the startd and endd are only updated on the mounted event, not when the props change.

PR #33, opened 2019-11-07
Rework group page routing/state to support back

nicksellen:
Not finished yet, there are 6 places that use the `GroupSelect` component that need to be reworked, I did 2 so far.

It was a bit of tangle of state with the values being read/set in both the page component and the `GroupSelect` component, was quite hard to reason about the ordering. The tricky bit with that architecture was there was no real way to tell from the page component when the group had changed because someone had actively selected one, or it was updating from the store on initial load.

So, I thought to make the `GroupSelect` more dumb. Now, it gets the list of groups from the store, but does not handle the remembering group feature. That is handled by a new mixin, to keep the page components simple too.

So, the new responsibility of the `GroupSelect` is just to implement a `v-model` interface for setting/getting the value of the group.

The mixin is where the brains are now, in one place, it's job is to check the route params, check the remember setting, and bind the correct group value for the `GroupSelect` to use.

It's a bit fiddly as there are a surprisingly large number of value/state combinations, but it's at least in one place! (was previously spread between each page component and the `GroupSelect` component.

I also added a `group/forget` action - so that when you select "all groups" (id=0) it doesn't save that, but clears the state. I think that more accurately reflects the intention.

Also, it's important to `unwatch()` watchers if they are created in a component, see https://forum.vuejs.org/t/vuex-unwatch-getter-is-this-right/68154/4 for details.

I'll add a few comments inline to explain how it works....

I would also quite like to add a test to recreate what I've been doing manually for now, as it could be easy to accidentality break.


components/GroupSelect.vue
```diff
@@ -11,37 +11,29 @@ select {
 <script>
 export default {
   props: {
+    value: {
```
nicksellen: This is how to implement a `v-model` inside a custom component, this and doing `$emit('input', newVal)` elsewhere.

mixins/createGroupRouteMixin.js
```diff
@@ -0,0 +1,69 @@
+/**
+ * A function to create a mixin to be used where there is an
+ * optional group param in the route.
+ *
+ * Designed to be used with the GroupSelect component.
+ *
+ * @param key used for the remember key and the route segment
+ * @returns a Vue mixin
+ */
+export default function createGroupRouteMixin(key) {
```
nicksellen: At the moment assumes the remember key and the route path are the same, e.g. `communityevents` is the remember key and `/communityevents/` is the path, but could added more args to implement that if needed.

mixins/createGroupRouteMixin.js
```diff
@@ -0,0 +1,69 @@
...
+  const rememberId = key
+  function routePath(id) {
+    return `/${key}/` + (id === 0 ? '' : id)
+  }
+  return {
+    computed: {
+      groupid: {
```
nicksellen: Anything using this mixin automatically gets a readable/writable `groupid` field.

mixins/createGroupRouteMixin.js
```diff
@@ -0,0 +1,69 @@
...
+              this.$store.commit('group/forget', {
+                id: rememberId
+              })
+            }
+          }
+        }
+      }
+    },
+    beforeDestroy() {
+      if (this._unwatchGroupRemember) this._unwatchGroupRemember()
```
nicksellen: Important! Or old instances get left around wrecking havoc and confusion!

See https://forum.vuejs.org/t/vuex-unwatch-getter-is-this-right/68154/4

mixins/createGroupRouteMixin.js
```diff
@@ -0,0 +1,69 @@
...
+    },
+    created() {
+      this._unwatchGroupRemember = this.$store.watch(
+        (state, getters) => getters['group/remembered'](rememberId),
+        (val, oldVal) => {
+          if (oldVal === undefined && val === undefined && this.groupid !== 0) {
```
nicksellen: This logic is a bit gnarly, and hard to understand I think....

Due to using `immediate: true` below, we get called _immediately_ (clever naming huh!) with the values, so we don't need to make a seperate call to get the remember value first....

mixins/createGroupRouteMixin.js
```diff
@@ -0,0 +1,69 @@
...
+            // Nothing set so far... make it what our current page is
+            this.$store.commit('group/remember', {
+              id: rememberId,
+              val: this.groupid
+            })
+          } else if (val !== undefined) {
+            if (this.groupid === 0) {
+              // We have a remember value, but we're on the general page
+              // Replace the current route
+              this.$router.replace(routePath(val))
```
nicksellen: This is the magic method to change the route without adding it into the back stack....

pages/communityevents/_id.vue
```diff
@@ -5,10 +5,11 @@
       <b-col cols="12" md="6" class="p-0">
         <div>
           <h1>Community Events</h1>
+          <pre>groupid: {{ groupid }}</pre>
```
nicksellen: Whoops, need to remove this.

pages/communityevents/_id.vue
```diff
@@ -63,31 +65,10 @@ export default {
...
-      // Ensure our select is set to the right value
-      this.$store.commit('group/remember', {
-        id: 'communityevents',
-        val: this.groupid
-      })
-    }
   },
-
   methods: {
-    groupChange: function(newGroup) {
```
nicksellen: This page component gets much simpler too, with all this logic outsourced to the mixin...

nicksellen:
Ok, thats me for today, there is a nice new component for wrapping the group select with memory.

Next bit is going through the remaining few places with a group select that needs either replacing with the wrapper wrapper component (volunteer and event modals), or doing the routing thing (stories)

components/CommunityEventModal.vue
```diff
@@ -101,7 +101,7 @@
             <label for="group">
               For which community?
             </label>
-            <groupSelect id="editevent" class="" @change="groupChange" />
+            <groupRememberSelect v-model="groupid" remember="editevent" />
```
nicksellen: Not sure if this works correctly when editing an existing event. I created one in the playground group, but it disappeared after pressing save, maybe into moderation?

components/CommunityEventModal.vue
```diff
@@ -326,17 +326,6 @@ export default {
 
       // If we don't have any groups, force a select.
       this.event.groups = this.event.groups ? this.event.groups : [{ id: 0 }]
```
nicksellen: Maybe this line can be removed too, I didn't quite understand why it's there...

mixins/createGroupRouteMixin.js
```diff
@@ -0,0 +1,89 @@
...
+            // We have changed the groupid away from the one in the route! Redirect...
+            this.$router.push(routePath(val))
+            if (val === 0) {
+              this.forgetGroup()
+            }
+          }
+        }
+      }
+    },
+    created() {
```
nicksellen: TODO: check how this is for server side stuff, as `created` is also called on the server...

edwh:
Thanks for this.
GroupSelect - Can you add a TODO DESIGN for the max-width setting in 
GroupSelect.  I think I put that hardcoded width in there but it can't 
be right.createGroupRouteMixinHas quite a lot of hardcoded 0s.  Constant 
for clarity?Sorry, nitpicking, but I'm not mad on the name having 
"create" and "mixin" in there.  The mixins/folder tells you it's a 
mixin, and all mixins would qualify to be called create, wouldn't they?  
For my education: When is it correct to set up a watcher inside 
created() rather than use the watch: option 
(https://vuejs.org/v2/guide/computed.html#Watchers)?  I'd rashly assumed 
that took care of destruction for you, but I'm probably missing 
something.Based on the previous point, can you (if you haven't already) 
check for any other watch leaks?StartEndCollection We removed all uses 
of index in keys because they are bad - see discussion with Jason.  But 
you've put this one back in explicitly.  Why's that?

nicksellen:
> Can you add a TODO DESIGN for the max-width setting in GroupSelect.

Yup!

> Sorry, nitpicking, but I'm not mad on the name having "create" and "mixin" in there.

Agree about the "*Mixin" suffix, will change that. This one returns a function that creates a mixin (to enable parameters to be passed in) rather than a mixin directly hence the _create_ prefix, makes it clearer that you need to call it first, not just use it directly. What do you think?

> createGroupRouteMixinHas quite a lot of hardcoded 0s.  Constant for clarity?

Yup, will do. would name it `ALL_GROUPS` or something... 0 was already used for that, so I hope there's not actually a group with id=0.

> For my education: When is it correct to set up a watcher inside created() rather than use the watch: option (https://vuejs.org/v2/guide/computed.html#Watchers)?  I'd rashly assumed that took care of destruction for you, but I'm probably missing something.

Using `watch:` option, the destruction is taken care of. You use that option to watch properties that are on your vue instance, in this case we're watching a getter on the store, so can't put the getter expression as a key in the `watch`: option. Having said that, you can declare a computed property that uses the getter, then watch that... that's what I used in the GroupRememberSelect iirc.

Something like:

```js
{
  computed: {
    foo() {
      return this.$store.getters['blah']
    }
  },
  watch: {
    foo(val) {
      console.log('foo changed!', val)
    }
  }
}
```

The more proper way actually for this is to use [mapGetters](https://vuex.vuejs.org/guide/getters.html#the-mapgetters-helper) from vuex to bring the getter value into a local property, but you have defined all the getters as functions, so can't use that. That's another topic for now ;)

```js
{
  computed: {
    ...mapGetters({
      foo: 'somemodule/somegetter'
    })
  }
}
```

> Based on the previous point, can you (if you haven't already) check for any other watch leaks?

Would do that in a seperate PR to keep this one focused.

> StartEndCollection We removed all uses of index in keys because they are bad - see discussion with Jason.  But you've put this one back in explicitly.  Why's that?

The replacement key was totally wrong as `date.toISOString` is not a function, it's an object with a start/end field, that can both be null, so it's not a valid key at all. I just reverted the change to avoid unrelated changes here, I don't know what it should use for the key.

edwh:

nicksellen:
> I see.  Just JSON encode it, that would work, I think?

That wouldn't work. They key is to describe a persistent identity of the item. Using the JSON encoded value would use the value as identity - but the value will change when you select date/time values.

I think it is broken using id (when I press remove and have two dates listed, the wrong one is going...), but even more broken when using "date.toISOString". Needs more thought/work than I think is appropriate here... which is actually meant to just be about the back button ;)

edwh:
You're right - mission creep.  Stick a TODO in there as part of your fix 
so we don't lose it.

components/CommunityEventModal.vue
edwh: Yes, and I later zapped it.  Create a new one if you want and I'll approve it.

components/CommunityEventModal.vue
edwh: When we first add an event, we won't have any groups.  We want to force them to choose one.  That might happen because of the order of the select options (as choose as at the top) but I see no harm in making it explicit.

edwh: We have also seen cases in the past where events get stranded without groups due to bugs and we want to edit them to fix them up.

nicksellen: I still don't really understand. It seems an event can have multiple groups (as I see the AddGroup/RemoveGroup stuff) but the UI only has a place for setting a single group?

singlegroup

edwh:
Half-baked future-proofing.  For messages, events, volunteer ops, they 
can all only have a single group at present.  But they may have multiple 
groups in future.

PR #36, opened 2019-11-08
Save state on page unload

nicksellen:
Addressing the TODO item:

> TODO Can we force it to happen before moving away from the page?  There's a window where if you sign in, then reload rapidly, you'll be logged out.  Similarly for other data changes.

This uses the "beforeunload" page event to add a hook to save the latest state.

### DIFF
--- a/plugins/vuex-persistedstate.js
+++ b/plugins/vuex-persistedstate.js
@@ -4,9 +4,10 @@ import createPersistedState from 'vuex-persistedstate'
 import cloneDeep from 'lodash.clonedeep'
 import requestIdleCallback from '~/assets/js/requestIdleCallback'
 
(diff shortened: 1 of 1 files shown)


edwh:
We make zillions of calls to alter state in the store (this is why I 
added the stuff to only save occasionally).  I'm slightly worried about 
the perf hit of creating a closure each time, and I'm not sure it's 
necessary to do that given that we have the settingState variable.  Can 
you have a think?

nicksellen:
I would not worry about that, I think a lot more is happening on each state change than the creation of a closure, however, I changed it anyway to reuse the `settingState` - but in order to access the storage/key to save under, instead of storing in a closure, they are now in top level const variables.

PR #37, opened 2019-11-11
Move notification specific classes into their component.

JayyajGH:
More clearing out of global styles.  Move the notification classes out of global and into the layout component.

Can the following element be removed?
`<a class="d-none dropdown-item" />`

Was it just added to sort out the previous `scoped` issue?  If so then I`ll remove it and re-test.

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -259,11 +259,6 @@ button {
     opacity: 0;
 }
 
-.notiflist .dropdown-menu {
-    height: 500px;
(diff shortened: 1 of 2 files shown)


edwh:
Yes, if you see items with d-none and a class that otherwise doesn't 
appear in that template, then it's because I didn't know about the deep 
selector stuff.  If you remove that one then I'll merge this.  Thanks.

JayyajGH:
That should now be removed.

PR #43, opened 2019-11-11
Refine GroupRememberSelect to work in edit forms

nicksellen:
This fixes the issue that the GroupRememberSelect would give too much importance to the remember value. When editing an existing event, it should use the value from the event.

This changes it so that the remember value is only used where the value is not set (i.e. falsey) on the outer component, so it means setting groupid to null in the pages that use this component is required to get the correct behaviour.

### DIFF
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -335,7 +335,7 @@ export default {
             }
           ]
 
-      if (this.event.groups) {
+      if (this.event.groups && this.event.groups.length > 0) {
         this.groupid = this.event.groups[0].id
       }
     },
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -296,7 +296,7 @@ export default {
     return {
       showModal: false,
       editing: false,
-      groupid: 0,
+      groupid: null,
       uploading: false,
       oldphoto: null,
       olddates: null,
@@ -335,8 +335,9 @@ export default {
             }
           ]
 
-      // If we don't have any groups, force a select.
-      this.event.groups = this.event.groups ? this.event.groups : [{ id: 0 }]
+      if (this.event.groups) {
+        this.groupid = this.event.groups[0].id
+      }
     },
     hide() {
       this.editing = false
--- a/components/GroupRememberSelect.vue
+++ b/components/GroupRememberSelect.vue
@@ -32,7 +32,7 @@ export default {
   },
   computed: {
     rememberedValue() {
-      return this.$store.getters['group/remembered'](this.remember) || 0
+      return this.$store.getters['group/remembered'](this.remember)
     },
     selectValue: {
       get() {
@@ -40,10 +40,6 @@ export default {
       },
       set(val) {
         val = parseInt(val)
-        // value changed from the select
-        if (this.rememberedValue !== val) {
-          this.updateMemory(val)
-        }
         if (this.value !== val) {
           this.$emit('input', val)
         }
@@ -54,14 +50,16 @@ export default {
     rememberedValue: {
       immediate: true,
       handler(val) {
-        // value changed from the memory!
-        if (this.value !== val) {
-          this.$emit('input', val)
+        // value received from memory (might be nothing)
+        // we only take it if there is not already a value
+        // this ensures we don't override explicitly set values from outside
+        if (!this.value) {
+          this.$emit('input', val || 0)
         }
       }
     },
     value(val) {
-      // value changed from outside
+      // value changed
       if (this.rememberedValue !== val) {
         this.updateMemory(val)
       }
@@ -69,14 +67,14 @@ export default {
   },
   methods: {
     updateMemory(val) {
-      if (val === 0) {
-        this.$store.commit('group/forget', {
-          id: this.remember
-        })
-      } else {
+      if (val) {
         this.$store.commit('group/remember', {
           id: this.remember,
           val
+        })
+      } else {
+        this.$store.commit('group/forget', {
+          id: this.remember
         })
       }
     }
--- a/components/GroupRememberSelect.vue
+++ b/components/GroupRememberSelect.vue
@@ -5,6 +5,11 @@
 </template>
 <script>
 import groupSelect from './GroupSelect'
+
+function intOrNull(val) {
+  return typeof val === 'number' ? parseInt(val) : null
+}
+
 export default {
   components: {
     groupSelect
@@ -39,7 +44,7 @@ export default {
         return this.value
       },
       set(val) {
-        val = parseInt(val)
+        val = intOrNull(val)
         if (this.value !== val) {
           this.$emit('input', val)
         }
@@ -50,12 +55,10 @@ export default {
     rememberedValue: {
       immediate: true,
       handler(val) {
-        // value received from memory (might be nothing)
+        if (val === undefined) return // no remembered value
         // we only take it if there is not already a value
         // this ensures we don't override explicitly set values from outside
-        if (!this.value) {
-          this.$emit('input', val || 0)
-        }
+        if (this.value === null) this.$emit('input', val)
       }
     },
     value(val) {
@@ -67,7 +70,7 @@ export default {
   },
   methods: {
     updateMemory(val) {
-      if (val) {
+      if (typeof val === 'number') {
         this.$store.commit('group/remember', {
           id: this.remember,
           val
--- a/components/GroupSelect.vue
+++ b/components/GroupSelect.vue
@@ -12,6 +12,16 @@ select {
 <script>
 export default {
   props: {
+    /**
+     * Selected value
+     *
+     *   if null
+     *     if   all=true       --> "all groups"
+     *     else all=false      --> "you must select a group"
+     *   else if it's a number --> use that group id
+     *
+     * (0 is not a valid group number)
+     */
     value: {
       type: Number,
       default: null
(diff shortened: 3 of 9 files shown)


edwh:
Are you sure you want to use falsey rather than an explicit test against 
null?  Wouldn't this mean that a remembered value would overwrite an 
explicitly passed value of 0 (All my groups)?

nicksellen:
> Are you sure you want to use falsey rather than an explicit test against null?

Note: it can also be undefined (in the case there is no remember value).

> Wouldn't this mean that a remembered value would overwrite an explicitly passed value of 0 (All my groups)?

I think it would, but I'm also not sure what this case means in practise, why would you want to do that? Seems like in that case the remember functionality is not really desired, no? (Also, remember I have added the ability to "forget" the remember value when setting to 0, so in the cases I have encountered undefined/null/0 are essentially the same, I think).

edwh:
Suppose you had a select which had:
a "please choose" option which goes at the top as the default if nothing 
is specifiedan "all my groups" option with a value of 0, which is just 
as valid to want to actively remember as a specific group is
There are ways round that, e.g. don't use 0 for all groups.  But 
generally over time I've found that I used to do loads of falsey tests 
in code, and now try to avoid them because I find they tend to lead to 
me not thinking clearly about the different options.  So I would prefer 
to avoid falsey tests in this code.  I wondered about it when I reviewed 
the forget code, but didn't comment at that time.

nicksellen:
> Suppose you had a select which had: a "please choose" option which goes at the top as the default if nothing is specifiedan "all my groups" option with a value of 0

I'm not sure I exactly get what you're saying, as the line formatting is a bit odd, but I _think_ you were saying the select might have a "please choose", then "all groups", then individual groups? But in the code, it can _either_ have a "please choose" _OR_ an "all groups", but not both:

https://github.com/Freegle/iznik-nuxt/blob/db674e912291a7d3fc6787f682000898350ab3a7/components/GroupSelect.vue#L44-L56

(out of interest, in the docs, they actually use "null" as the value for the "please choose" item https://bootstrap-vue.js.org/docs/components/form-select/)

So, given this, what does remembering a value of 0 mean? And in which places should there be different behaviour compared to not having a value remembered?

I tried out all the places where a group select is used to see if the behaviour made sense, and from what I could tell it does, but I could be missing a case, but I don't know what.

I agree in general about falsey and not thinking about the options. I could add in some more explicit checks for ALL_GROUPS, although also have to deal with the possibility of null aswell as undefined. I think first we need to resolve the above bit though.

edwh:
On 11 November 2019 20:08:34 GMT, Nick Sellen <notifications@github.com> wrote:
>> Suppose you had a select which had: a "please choose" option which
>goes at the top as the default if nothing is specifiedan "all my
>groups" option with a value of 0
>
>I'm not sure I exactly get what you're saying, as the line formatting
>is a bit odd, but I _think_ you were saying the select might have a
>"please choose", then "all groups", then individual groups? But in the
>code, it can _either_ have a "please choose" _OR_ an "all groups", but
>not both:
>

It's a thought experiment to illustrate the dangers of relying on falsey tests, and why IMO they are a bad code habit to have. 

Just because the code currently happens to work in a way in which that bad habit wouldn't cause actual bugs is not a good reason to have them in there. 

I want to avoid falsey tests whether or not they would currently cause problems. It makes the code more robust against future bugs.

nicksellen:
At the moment we have to deal with 4 possible types of value `groupid`:
- `null` - default or initial value set in some places
- `0` - means either "all groups" (if that option is enabled), or "please choose a group"
- `Number > 0` - means the id of a group
- `undefined` - when there is no remember value saved, and we use an "immediate: true" watcher (more of an implementation detail, not a meaningful value of groupid)

I (mostly) agreed about falsey tests, I don't think we need to discuss it further in the abstract. My questions are more about the intended behaviour and the data model, than coding practise.

I would like to find out about behaviour differences expected in these different cases. From what I've seen/tried myself the following should be true:
- if groupid is `Number > 0` that group should be selected
- any other value should show either "all groups" or "please choose a group"

After seeing in the code that `0` means either "all" or "choose", I deduced that you _cannot_ have a select that has:

- "please choose"
- "all groups"
- "example group one"
- "example group two"

... and therefore that I can code according to that assumption.

But you also said it's valid to actively remember an "all groups" setting, so I think maybe I'm missing a use case, but then again I don't think there would be any different behaviour when not actively remembering "all groups", so if it _is_ needed there might be a bug there.

What I am getting at, is that I need to know what _should_ be the behaviour of the component, like an informal specification. So far I based it on the code, and by trying out the site to see what makes sense in each use case I can find.

From what I think/understand/suspect - we could actually just have:
- `null` - means either "all groups" (if that option is enabled), or "please choose a group" (follows the [bootstrap example](https://bootstrap-vue.js.org/docs/components/form-select/) and semantically feels accurate to me - "I am not filtering by group, and if I a group is required, I need to choose, otherwise all of them")
- `Number` - means the id of a group

So, two questions:
1) what would remembering a value of `0` mean, as in, what would behave differently for a user? I don't think I found a case where it would make a difference, hence adding a "forget" option for that case
2) what is the difference in behaviour between a value of `null` and a value of `0`? I don't think I found a case where there is a difference

Sorry it's a bit long!

edwh:
Yes, it's a bit confusing to reverse-engineer the desired behaviour from 
the code, isn't it?  Thanks for perservering.

At the moment you are correct, that we never in practice have both 
"please choose" and "all my groups" - we decide which based on the "app" 
prop, and they are mutually exclusive.  Therefore there would be no 
difference in practice between remembering a value of 0, remembering a 
value of null, or not remembering any value at all.

However because we are persisting this information, it's prudent to 
consider whether what we do now will introduce migration issues later.

In future we might want the ability to have both "please choose" and 
"all my groups".  I have an example in mind.  Details omitted here (ask 
if you like).

When we add that ability, we will potentially have migration issues if 
we remember both "please choose" and "all my groups" in the same way in 
the persistent store.  Reasoning omitted here (ask if you like).

Because of that, I would prefer to distinguish between "please choose" 
and "all my groups" in what is persisted in the store, even though we do 
not need to do so for the current function.

So the answer to both your questions is that there is no difference now; 
this is future-proofing.

Does that help?  I'm sure I wasn't clear previously that I was talking 
about potential future function here, so that will have been confusing.

nicksellen:
So, I make a proposal with the goal:
- keep behaviour as I have understood/described it
- if/when you want to introduce the ability to have both "all groups" and "choose" options available, existing remembered value will not have a migration issue

Allow two types of value:
- `null` - means either "all groups" (if that option is enabled), or "please choose a group"
- `Number` - means the id of a group

It should remember value when it is `Number`, forget value when it is `null`. If remembered value is not a valid option, set to `null` and forget (this is an improvement on current behaviour, as right now you could have a remembered group, then leave it, and end up with a select value for which there is no option).

When it comes to adding the ability to have both "all groups" and "choose", `0` is added as another valid option to mean "all groups". Then, the same remember rule applies (as `Number` includes `0`).

The main principle behind it is that the select always has a default value if nothing explicitly set, and so we don't need to remember the value if it the same as the default, only changes from it. Keeping as little in remembered state as possible - no migration issues when you don't have state to migrate, and slightly reduces state loading complexity for default cases.

edwh:
Ok, that sounds fine.

nicksellen:
Done! There are quite a lot of cases, and all the testing is manual, so hard to be sure I've captured all the cases, but I _think_ it's OK. Ready for review.

I found an error case whilst doing it too, see https://github.com/Freegle/iznik-nuxt/pull/43/commits/d5fa15bc094e77dd183d1323b9bf021173e10ec9 (`ret.status === 200` not being a way to check for error state from the Freegle API). But that's for another time/PR.

edwh:
Thanks, I'll have a look tomorrow morning.  Away over the weekend, so 
not much from me.

Good spot on the ret.status.  It was a mistake when I wrote the server 
API not to return the error/success via HTTP codes, and I have tripped 
up a couple of times when calling them here.  Make sure you add a TODO 
if you've not already for anything like that which you spot and aren't 
fixing in the current PR.

nicksellen:
Yup, I picked up the "TODO's as issue tracking" pattern by now ;) - I added two in the last changes (I broken the `StartEndCollection` on the volunteer page, as it doesn't use `v-model` yet).

edwh:
In CommunityEventModal and VolunteerOpportunityModal you have an if test 
on groups.  I believe an empty array is truthy, and so it would be 
better to test both on the array and the length to avoid an error if the 
array is present but empty.  That's definitely necessary for Volunteer 
Ops, where we do have system wide opportunities which have no group.

In GroupRememberSelect, in updateMemory, I wondered about the case where 
val = null.  I think that can't happen because we only call this method 
when the select updates.  But maybe add a comment to that effect to 
avoid having to think about it.  There's similar code in 
createGroupRoute which might be worth a comment too.

In GroupSelect, thanks for simplifying the loop.  I seem to have a 
mental block about the various ways of iterating through things and 
often end up with a clunky method.  Good to catch the invalid case.

nicksellen:
Yup, good point about the array check, I'll fix that. I think I just borrowed the existing logic, which also didn't check the length, which used to be:

```js
// If we don't have any groups, force a select.
this.event.groups = this.event.groups ? this.event.groups : [{ id: 0 }]
```

> In GroupRememberSelect, in updateMemory, I wondered about the case where val = null.  I think that can't happen because we only call this method when the select updates.

As per the discussion above, `null` means "please select" or "all groups" (depending on the `all` option), and is the default, so `null` will be set when you choose that option.

PR #44, opened 2019-11-11
Fix saving/editing community event dates

nicksellen:
The previous code would make a huge number of `AddDate` API calls. I _think_ this was actually because there was some `JSON.stringify()` calls being done on the start/end date object, which it was then iterating over. Not sure... I didn't look too deeply into that.

I had also noticed that the "event" data on the `CommunityEventModal` was not reactive (i.e. trying to print or watch values did not work). This was because the event object was defined as `{}` with no attributes, but used in the forms like `event.title`. This does not allow vue reactivity to work. So, the new approach is to define a default value function that gives null values to all the attributes that be used.

I reworked it to use `v-model` attributes where possible, this results in quite a lot of code/logic going away as we don't need manual change logic so much when we work directly on the values with v-model.

I would also propose to rework the `uniqueid/generate` implementation, I notice that existing dates (i.e. returned from the db, already have their database id, so we can use that as the key, so it's only for new entries that have yet to be saved. In the new approach, all that logic is contained within `StartEndCollection` so we can simply have a counter in that file (and make the keys like "new-1", "new-2", etc... ). The scope of the id is only for that single file, so no need to keep it in the global store.

... but I would wait for input from @edwh to go ahead with that.

... also, at some point would like to tackle validation and other concerns, lets see...!

### DIFF
--- a/components/StartEndCollection.vue
+++ b/components/StartEndCollection.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <div v-for="(date, idx) in value" :key="date.uniqueid" :class="date.string && date.string.past ? 'inpast': ''">
-      <StartEndDate v-model="value[idx]" @remove="remove(idx)" />
+      <StartEndDate v-model="value[idx]" @remove="remove(date)" />
     </div>
     <b-btn variant="white" class="mt-1" @click="add">
       <v-icon name="plus" /> Add another date
@@ -44,8 +44,9 @@ export default {
     }
   },
   methods: {
-    remove(idx) {
-      this.value.splice(idx, 1)
+    remove(item) {
+      const idx = this.value.indexOf(item)
+      if (idx !== -1) this.value.splice(idx, 1)
     },
     async add() {
       this.value.push({
--- a/components/StartEndCollection.vue
+++ b/components/StartEndCollection.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
-    <div v-for="date in editableDates" :key="'startend-' + date.uniqueid" :class="date.string && date.string.past ? 'inpast': ''">
-      <StartEndDate :start="date.start" :end="date.end" :uniqueid="date.uniqueid" @remove="remove(date.uniqueid)" @change="change(date)" />
+    <div v-for="(date, idx) in value" :key="date.uniqueid" :class="date.string && date.string.past ? 'inpast': ''">
+      <StartEndDate v-model="value[idx]" @remove="remove(idx)" />
     </div>
     <b-btn variant="white" class="mt-1" @click="add">
       <v-icon name="plus" /> Add another date
@@ -21,53 +21,39 @@
 <script>
 // TODO Validation - end date > start date, no stupidly long events (3 days?), no overlapping dates, only one blank slot.
 // TODO This code could do with a bit of extra testing - various combinations of add/edit/cancel/delete etc.
-const StartEndDate = () => import('~/components/StartEndDate')
+import StartEndDate from '~/components/StartEndDate'
 
 export default {
   components: {
     StartEndDate
   },
   props: {
-    dates: {
+    value: {
       type: Array,
       required: true
     }
   },
-  data: function() {
-    return {
-      editableDates: []
-    }
-  },
-  mounted: function() {
-    this.editableDates = this.dates ? this.dates : []
-  },
-  methods: {
-    remove(uniqueid) {
-      for (let i = 0; i < this.editableDates.length; i++) {
-        if (this.editableDates[i].uniqueid === uniqueid) {
-          this.editableDates.splice(i, 1)
-        }
-      }
-
-      this.$emit('change', JSON.stringify(this.editableDates))
-    },
-    add() {
-      this.editableDates.push({
-        uniqueid: this.$store.dispatch('uniqueid/generate'),
+  async mounted() {
+    if (this.value.length === 0) {
+      this.value.push({
+        uniqueid: await this.$store.dispatch('uniqueid/generate'),
         start: null,
         end: null,
         past: false
       })
+    }
+  },
+  methods: {
+    remove(idx) {
+      this.value.splice(idx, 1)
     },
-    change(date) {
-      for (let i = 0; i < this.editableDates.length; i++) {
-        console.log('Check', i)
-        if (this.editableDates[i].uniqueid === date.uniqueid) {
-          this.editableDates[i] = date
-        }
-      }
-
-      this.$emit('change', JSON.stringify(this.editableDates))
+    async add() {
+      this.value.push({
+        uniqueid: await this.$store.dispatch('uniqueid/generate'),
+        start: null,
+        end: null,
+        past: false
+      })
     }
   }
 }
--- a/components/StartEndDate.vue
+++ b/components/StartEndDate.vue
@@ -8,7 +8,7 @@
         <!-- Add form-control class to get focus etc. -->
         <date-picker
           id="startDate"
-          v-model="startd"
+          v-model="value.start"
           class=""
           lang="en"
           type="datetime"
@@ -16,14 +16,13 @@
           format="ddd, Do MMM HH:mm a"
           :time-picker-options="{ start: '00:00', step: '00:30', end: '23:30' }"
           placeholder=""
-          @change="change"
         />
       </div>
       <div class="mr-lg-4 d-flex flex-column">
         <label for="endDate" class="date__label">Ends at:</label>
         <date-picker
           id="endDate"
-          v-model="endd"
+          v-model="value.end"
           class=""
           lang="en"
           type="datetime"
@@ -31,13 +30,12 @@
           format="ddd, Do MMM HH:mm a"
           :time-picker-options="{ start: '00:00', step: '00:30', end: '23:30' }"
           placeholder=""
-          @change="change"
         />
       </div>
     </div>
     <div>
       <!-- TODO Hide button if only one date present -->
-      <b-btn variant="white" size="sm" @click="$emit('remove', uniqueid)">
+      <b-btn variant="white" size="sm" @click="$emit('remove')">
         <v-icon name="trash-alt" title="Delete this date" aria-hidden="true" />
         <span class="delete__label">Remove</span>
       </b-btn>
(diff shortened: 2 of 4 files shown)


components/StartEndCollection.vue
```diff
@@ -21,53 +21,39 @@
 <script>
 // TODO Validation - end date > start date, no stupidly long events (3 days?), no overlapping dates, only one blank slot.
 // TODO This code could do with a bit of extra testing - various combinations of add/edit/cancel/delete etc.
-const StartEndDate = () => import('~/components/StartEndDate')
+import StartEndDate from '~/components/StartEndDate'
```
nicksellen: We don't need dynamic imports for _everything_ especially in cases like this, where `StartEndDate` is only used in this component.

components/StartEndCollection.vue
```diff
@@ -21,53 +21,39 @@
...
-    remove(uniqueid) {
-      for (let i = 0; i < this.editableDates.length; i++) {
-        if (this.editableDates[i].uniqueid === uniqueid) {
-          this.editableDates.splice(i, 1)
-        }
-      }
-
-      this.$emit('change', JSON.stringify(this.editableDates))
+    remove(idx) {
+      this.value.splice(idx, 1)
```
nicksellen: we don't need to use that uniqueid everywhere, in this case, the array index is perfectly sufficient

components/StartEndDate.vue
```diff
@@ -48,36 +46,9 @@
...
 export default {
   props: {
-    uniqueid: {
-      type: String,
+    value: {
+      type: Object,
       required: true
-    },
-    start: {
-      validator: prop => typeof prop === 'string' || prop === null,
```
nicksellen: I don't like these validator functions, it basically is undoing the work of `required: true` if null is allowed...

edwh:
The use of default to make reactivity work is useful - I think I added datesChange without digging too deeply, and this is a much better solution.  You imply this fixes the issue with AddDate calls, which is good.

I'm not sure about using the database id rather than uniqueid.  I started down that track when I added it, but we would still have to invent ids when adding/editing.  We would have to ensure that any ids we invented on the client side for new dates during edit didn't clash with ids which had been created by the server, and that seemed to be leading down an unpleasant path.  We are inventing these ids purely for the convenience of the client (proof: we didn't have them in the old client), and I don't think they should have any overlap with the server ids at all.  Stick with what it does.

I agree we don't always need dynamic imports.  There are some cases where we actively need static imports (e.g. to avoid screen flicker).

I would definitely prefer to continue to remove by unique id rather go back to doing it by index.  We discovered that indexes were more fragile than we thought, and there are likely to be other bugs in that area.  Suppose (for example) that you click twice in quick succession on the delete buttons for two different dates; we would have to rely on those events being processed in the same order to avoid the potential of deleting the wrong date, and whether or not this is what Vue currently does, it's the kind of behaviour that it's risky to rely on.  Whether or not that's a bug that actually exists, it's the _kind_ of bug that will tend to exist if we use indexes.  Using uniqueid just knocks those on the head.

The question about validators like this:

> validator: prop => typeof prop === 'string' || prop === null,

...is interesting.  I _think_ this means that you have to explicitly pass the property, even if the value is null, and I think that is different from allowing you to omit the property (which could mean you just forgot or didn't think about it).  That's my intention, anyway.

nicksellen:
> We would have to ensure that any ids we invented on the client side for new dates during edit didn't clash with ids which had been created by the server, and that seemed to be leading down an unpleasant path.

I can be sure that "new-1" will _never_ be a database id. But even if we just wanted unique ids on client, a vuex state/action is overkill imo. It's not a global concern of the dates arrays, it's just about that component.

> I would definitely prefer to continue to remove by unique id rather go back to doing it by index. We discovered that indexes were more fragile than we thought, and there are likely to be other bugs in that area.

Using the index was not suitable for a :key when editing a form, that was about keeping the relation with the component and item in the array. Using the index to remove an item is standard, see:
- https://vuejs.org/v2/guide/list.html#v-for-with-a-Component
- https://stackoverflow.com/a/41374441
- https://stackoverflow.com/a/42845755
- https://medium.com/vuejs-tips/tip-3-deleting-array-index-in-vue-js-8cc31d7214bf

edwh:
On 11 November 2019 19:42:31 GMT, Nick Sellen <notifications@github.com> wrote:
>> We would have to ensure that any ids we invented on the client side
>for new dates during edit didn't clash with ids which had been created
>by the server, and that seemed to be leading down an unpleasant path.
>
>I can be sure that "new-1" will _never_ be a database id. 

Sure. But it's architecturally wrong for the client to invent values of something that belongs to the server. 

>But even if
>we just wanted unique ids on client, a vuex state/action is overkill
>imo. It's not a global concern of the dates arrays, it's just about
>that component.
>

We may need unique ids elsewhere, not just to do with dates. 

>> I would definitely prefer to continue to remove by unique id rather
>go back to doing it by index. We discovered that indexes were more
>fragile than we thought, and there are likely to be other bugs in that
>area.
>
>Using the index was not suitable for a :key when editing a form, that
>was about keeping the relation with the component and item in the
>array. Using the index to remove an item is standard, see:
>- https://vuejs.org/v2/guide/list.html#v-for-with-a-Component
>- https://stackoverflow.com/a/41374441
>- https://stackoverflow.com/a/42845755
>-
>https://medium.com/vuejs-tips/tip-3-deleting-array-index-in-vue-js-8cc31d7214bf

If the index is passed on the event, then presumably it is a closure and the value is captured at the time of list rendering?

So if we have two remove events triggered in very quick succession, then we need to be confident that  there is no timing window such as:
- trigger remove for A
- trigger remove for B
- queue callback for remove for A at index 0
- queue callback for remove for B at index 1
- execute remove for index 0
- execute remove for index 1 - which would be wrong 

This is pretty subtle behaviour. It may well be that Vue is carefully enough coded that we can simply trust that they have considered this and the event queue is processed in a way that means it can't happen, and that no such bugs will ever be introduced. 

Why would we depend on that, though? The fact we're having this time-consuming discussion just shows it requires too much thought, doesn't it? 

Remove by unique id is robust against those kinds of bugs, and transparently so. It means you don't have to have subtle conversations like this. Stick with it, please.

edwh:
For when you come back to this (not today)...

I created a hacky test:  
* There's a button on the Add Event modal to set up three test dates
* There's another button which triggers a click on the first two delete buttons in quick succession

Slightly to my surprise, there does seem to be an actual issue there - this deletes the first and third dates leaving the second, whereas I would expect it to delete the first two dates leaving the third.  

Alternatively, this indicates something I have failed to understand.  Both are possible.

This might explain more clearly the kind of problem which I think means that using index is best avoided.  I may be wrong about all of this, though, so I would be interested to see what you think.

Here's a patch file:

```
Index: components/StartEndCollection.vue
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- components/StartEndCollection.vue	(date 1573493555000)
+++ components/StartEndCollection.vue	(date 1573554180569)
@@ -1,5 +1,11 @@
 <template>
   <div>
+    <b-btn variant="info" @click="testSetup">
+      Test setup
+    </b-btn>
+    <b-btn variant="info" @click="testRemove">
+      Test remove
+    </b-btn>
     <div v-for="(date, idx) in value" :key="date.uniqueid" :class="date.string && date.string.past ? 'inpast': ''">
       <StartEndDate v-model="value[idx]" @remove="remove(idx)" />
     </div>
@@ -21,6 +27,7 @@
 <script>
 // TODO Validation - end date > start date, no stupidly long events (3 days?), no overlapping dates, only one blank slot.
 // TODO This code could do with a bit of extra testing - various combinations of add/edit/cancel/delete etc.
+import dayjs from 'dayjs'
 import StartEndDate from '~/components/StartEndDate'
 
 export default {
@@ -54,6 +61,38 @@
         end: null,
         past: false
       })
+    },
+    async testSetup() {
+      const now = dayjs()
+      this.value = []
+      this.value.push({
+        uniqueid: await this.$store.dispatch('uniqueid/generate'),
+        start: now,
+        end: now.add(1, 'day'),
+        past: false
+      })
+      this.value.push({
+        uniqueid: await this.$store.dispatch('uniqueid/generate'),
+        start: now.add(2, 'day'),
+        end: now.add(3, 'day'),
+        past: false
+      })
+      this.value.push({
+        uniqueid: await this.$store.dispatch('uniqueid/generate'),
+        start: now.add(4, 'day'),
+        end: now.add(5, 'day'),
+        past: false
+      })
+    },
+    testRemove() {
+      // Get the elements first, because then we definitely have the buttons corresponding to the first and second
+      // dates.
+      const butt0 = document.getElementById('remove0')
+      const butt1 = document.getElementById('remove1')
+
+      // Now delete the first and second.  We should be left with the third, but we are left with the second.
+      butt0.click()
+      butt1.click()
     }
   }
 }
Index: components/StartEndDate.vue
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- components/StartEndDate.vue	(date 1573493555000)
+++ components/StartEndDate.vue	(date 1573553258039)
@@ -35,7 +35,7 @@
     </div>
     <div>
       <!-- TODO Hide button if only one date present -->
-      <b-btn variant="white" size="sm" @click="$emit('remove')">
+      <b-btn :id="'remove' + index" variant="white" size="sm" @click="$emit('remove')">
         <v-icon name="trash-alt" title="Delete this date" aria-hidden="true" />
         <span class="delete__label">Remove</span>
       </b-btn>
@@ -49,6 +49,10 @@
     value: {
       type: Object,
       required: true
+    },
+    index: {
+      type: Number,
+      required: true
     }
   }
 }

```

nicksellen:
That is a pathological test case that a user would not trigger. However, I have changed it to use `indexOf` (object identity), that should pass your test case.

I will leave the uniqueid generation topic aside, as this PR does not touch that. There are also plenty of other niggles with the event date stuff, this PR does not address those, but just gets it into a _basically working_ state (you can create/edit events).

edwh:
Thanks, that definitely looks considerably neater code.

Do you want to add any TODOs for other niggles before I merge?

nicksellen:
Nope, no more TODOs, they are already in the code I think. Merge if you're happy :)

PR #45, opened 2019-11-13
Enhancements/add image carousel

JayyajGH:
More work to remove some global classes.  Things to note:
- The slide property used to set the alt text wasn`t working before.  I`ve simplified it and just used the loop index instead.
- More tidying can be done in this area but I`ll leave that for another PR.

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -247,31 +247,6 @@ nav .dropdown-toggle {
   border: none;
 }
 
-.carousel-control-prev-icon,
-.carousel-control-next-icon {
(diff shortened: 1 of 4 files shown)


JayyajGH:
Potential future change - MyMessage will always display a carousel even if there is only one image. Message will display a single image if there is only one.  Should this be made the same?

edwh:
Yes, that's better as a separate component.

Showing a single image if there is only one is what we should do in both cases.

JayyajGH:
I`ll tweak the component to add the single image bit in as well

PR #48, opened 2019-11-13
Enhancements/create notice component

JayyajGH:
This is the first stab at a notice component.

There are a couple of places I haven't updated yet where there are a few other things that need sorting first. In one location a dropdown list appears over my message but it is just hidden behind the message previously.  I think it needs completely hiding for example.

In one place I have changed the notice to a regular div as it was just a container for some other dropdowns and nothing to do with a notification message.

### DIFF
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -27,13 +27,13 @@ $color-blue--lighter: #2F9AF7;
 $color-blue--light: #4895DD;
 $color-blue--base: #0000FF;
 $color-blue--dark: #35495E;
+$color-blue--1: #0C5460;
 
 $color-yellow--light: #F9F7EC;
 $color-yellow--base: #F3F0DA;
 $color-yellow--dark: #FBE8AA;
-
-$color-yellow-1: #fff3cd;
-$color-yellow-2: #ffeeba;
+$color-yellow-1: #FFF3CD;
+$color-yellow-2: #FFEEBA;
 $color-yellow-3: #856404;
 
 $color-red: #FF0000;
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -32,6 +32,10 @@ $color-yellow--light: #F9F7EC;
 $color-yellow--base: #F3F0DA;
 $color-yellow--dark: #FBE8AA;
 
+$color-yellow-1: #fff3cd;
+$color-yellow-2: #ffeeba;
+$color-yellow-3: #856404;
+
 $color-red: #FF0000;
 
 $color-white: #FFF;
--- a/components/ChatMessageInterested.vue
+++ b/components/ChatMessageInterested.vue
@@ -30,8 +30,7 @@
               </h4>
             </b-card-title>
             <b-card-text>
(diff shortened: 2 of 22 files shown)


edwh:
Had a quick skim.  Generally looks fine.

Please ensure there are TODOs for the places you've not updated yet if 
there aren't already.

I wonder if we should pass the semantic meaning and let the component 
decide what style that should be, rather than add the background and 
border in the parent?

JayyajGH:
TODOs added...  I'll get to them next.

JayyajGH:
I was thinking about that too.  Maybe wait until I have finished all of the b-alert removals and have a think.  Some of the ones left are a different style. I've never found a way that I'm happy with to pass in information like that.  A string always seems a bit messy e.g. a messageType prop of "warning" or "error".  Having separate props as bools is not nice as you need to deal with someone potentially passing in two.  I guess the bootstrap b-alert passes in variant="info" etc.  I'll take a look.

edwh:
variant= is the standard Bootstrap Vue way to do it, so that would be 
the pattern to follow, I think.

JayyajGH:
I'll add it to the list...

JayyajGH:
Might be a few days before I get around to it though.

edwh:
That's absolutely fine.  What with you being a volunteer and all.

JayyajGH:
I couldn't help myself.  I won't get paid otherwise. :-p

JayyajGH:
@edwh - Could you take a look at these changes when you get a chance and see if you're happy with them now?  The others I've added todos for as each of them has their own complications which I'll look into next.  At least we can get the guts of it merged to save the branch going stale.  I'll add the other variants when I update those components.

edwh:
I think you might still have some classes in the parent, e.g. in 
ProfileModal you have class="bg-warning border-warning" .  So maybe 
search for notice-message.*bg to check?

Also check for "notice box" and "notice component" to make they're all 
gone.

After that I'll merge in.

JayyajGH:
Good spot!  The line was too long to show on my search before so I missed it! (That logic probably needs putting in a computed property..).  I'll update it tomorrow as I won't be able to test it whilst I'm on my work machine.

What are notice box and notice component?

edwh:
They're just words I stuck in some TODOs for when we had this component. 
  Just want to make sure they've all got caught in your branch.

JayyajGH:
Ah ok cool.  I was just searching for b-alert before so I'll search for them too.  One of the b-alerts I think is still correct.  The message that pops up when you enter your phone number so I'll leave that one as it is,

edwh:
That alert would be there next time you visited the page, though.  So 
maybe it needs to be an alert when you first add a number, and then a 
notice thereafter?

JayyajGH:
This is why accessibility is so hard!  I'll add a TODO for that one I think.

edwh:
Fair enough.  Let me know when you're ready for me to merge.

JayyajGH:
@edwh Right I think the easy ones are done.  There are 2 or 3 more complicated ones left which I will get to next...

edwh:
Ok, shall I merge?

JayyajGH:
Yeah go on.  The other stuff might take a while to sort...

PR #49, opened 2019-11-15
Create a computed method to check if the user has reneged

JayyajGH:
This is currently done in a few places.  Maybe we could add a function in the user store to do this?

### DIFF
--- a/components/ChatPane.vue
+++ b/components/ChatPane.vue
@@ -70,7 +70,7 @@
           </b-row>
           <b-row>
             <b-col class="p-0">
-              <notice-message v-if="warnuser" variant="warning">
+              <notice-message v-if="userHasReneged" variant="warning">
                 <v-icon name="exclamation-triangle" />&nbsp;Things haven't always worked out for this freegler.  That might not be their fault, but please make very clear arrangements.
               </notice-message>
               <!-- TODO Make this a notice-message component too -->
@@ -307,21 +307,8 @@ export default {
 
       return ret
     },
-
-    warnuser() {
-      let ret = false
-      if (this.otheruser) {
-        const user = this.$store.getters['user/get'](this.otheruser.id)
-        if (user && user.info) {
-          const info = user.info
-          ret =
-            info.reneged &&
-            info.reneged > 1 &&
-            (info.reneged * 100) / (info.reneged + info.collected) > 25
-        }
-      }
-
-      return ret
+    userHasReneged() {
+      return this.$store.getters['user/userHasReneged'](this.otheruser.id)
     },
 
     spammer() {
--- a/components/ChatPane.vue
+++ b/components/ChatPane.vue
@@ -308,7 +308,9 @@ export default {
       return ret
     },
     userHasReneged() {
-      return this.$store.getters['user/userHasReneged'](this.otheruser.id)
+      return this.otheruser
+        ? this.$store.getters['user/userHasReneged'](this.otheruser.id)
+        : false
     },
 
     spammer() {
--- a/store/user.js
+++ b/store/user.js
@@ -6,6 +6,20 @@ export const state = () => ({
   list: {}
 })
 
+function getUserByID(state, id) {
+  let user = null
+
+  Object.keys(state.list).forEach(key => {
+    const item = state.list[key]
+
+    if (parseInt(key) === parseInt(id)) {
+      user = item
+    }
+  })
+
+  return user
+}
+
 export const mutations = {
   add(state, item) {
     Vue.set(state.list, item.id, item)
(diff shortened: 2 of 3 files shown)


edwh:
I've added attributes like this in the store elsewhere, e.g. addFields in communityevents comes to mind.  

It's probably worth checking whether there is a "right" thing to do in Vue here, so I'm tagging @nicksellen .  


Options I can think of:

- Computations in templates, which are probably better as...
- ...computed properties in components, which could be shared by extending components or mixins
- Or move them into the store

E.g. some rule of thumb guidance of this form:

- If a computation is only used once, then do it in a computed property in the component.  
- If it's used more than once, move it into a mixin.
- If it's used more than thrice and it's pretty cheap, do it in the store then it's accessible everywhere if you need it.

nicksellen:
I would put it in the store in this case, as another getter, seems a perfect use case for it.

In karrot we actually augment most of the objects with useful properties in the getter, so it would be available as `user.reneged`, no seperate call to getters. But that pattern isn't in use in iznik, so I wouldn't add it now...

edwh:
Ok, getter it is.  @JayyajGH , can you do that?

JayyajGH:
I've used mapGetters from vuex in the past which creates computed properties for you.  I'll take a look at adding it into the store.  Need to work out how the user object works first.  Gonna be doing things slowly at the moment between other work and packing up the house!

nicksellen:
At the moment all the getters return functions, so we can't use `mapGetters`. They should not all be functions, but that's just the current codebase for now :)

JayyajGH:
So just to be clear, it looks like the user store contains a list of all relevant users. So in a chat it will contain both users, in a profile view it will contain a single user.  Have I got that right?  So basically we need to pass in the id to get the specific user we are interested in.

edwh:
Why shouldn't they be functions?

edwh:
Yes, you can assume that the store contains at least the users you are interested in.

JayyajGH:
What about something like this?  I still need to replace the logic in the other location too to use the store too.

components/ChatPane.vue
```diff
@@ -307,21 +307,8 @@ export default {
...
-          ret =
-            info.reneged &&
-            info.reneged > 1 &&
-            (info.reneged * 100) / (info.reneged + info.collected) > 25
-        }
-      }
-
-      return ret
+    userHasReneged() {
+      return this.$store.getters['user/userHasReneged'](this.otheruser.id)
```
nicksellen: This is missing the `if (this.otheruser)` check that the previous code had, so potentially `this.otheruser.id` would throw an error.

store/user.js
```diff
@@ -6,6 +6,20 @@ export const state = () => ({
...
+    const item = state.list[key]
+
+    if (parseInt(key) === parseInt(id)) {
+      user = item
+    }
+  })
+
+  return user
+}
+
```
nicksellen: This loop could be simplified as `state.list` is an object keyed by the id, so should be the same as:
```js
function getUserByID(state, id) {
  return state.list[id]
}
```
Try it first though...

components/ChatPane.vue
JayyajGH: Actually that check would probably be better in the getter maybe.

JayyajGH: Ah no.  You mean for id...

JayyajGH: Sorted...

store/user.js
JayyajGH: I decided not to touch this from what it was before.  But happy to update if everyone agrees.  It certainly looks a bit (much) nicer.

nicksellen: We can let @edwh decide :smile:

JayyajGH: Well it seems to work ok but I'll give it a proper test on Monday.  It is much more readable!

edwh:
I'm in favour of that. It is horrible as it was, and pointlessly so.

JayyajGH:
Ok I've now checked in an update to simplify that method.  Not sure if people like the pattern of doing all the param error checks first up on entering the function and returning asap.  I love this style as I think it simplifies the code block and removes the need for any nesting.  I can always change it if people hate it.

edwh:
I'm fine with returns at the start and end of a method; I don't like 
them in the middle.

Looks good, will merge.

JayyajGH:
Thanks!    Some people I know are still in the "only one return per method" camp which is why I asked.

edwh:
I have that prejudice, but I don't think it's justified, so I see no 
reason to inflict it on others.

JayyajGH:
Ha!  I've always had it too coming from an old-school C/C++ background but trying to get over it.

edwh:
That's why I have it, but I don't really see why it was true in C 
either.

PR #50, opened 2019-11-15
Fix start end collection usage

nicksellen:
The main change is to correctly use `v-model` when using `StartEndCollection`. I had done this for community events, but omitted to do it for volunteerings.

Other related changes are:
- create default event for volunteering with all the fields nulled (allows reactivity to work as vue intended, matches how it works for community events)
- add a few extra fields to the default function for events and volunteerings (id and canmodify), probably not important, as they are only set for data that comes back from the API I think, but defining the fields up front seems nice to me (specifies what is available)
- use lodash's `cloneDeep` instead of `JSON.parse(JSON.stringify(data))` (which has data loss when using it with dates, which we do, see [explanation](https://stackoverflow.com/a/122704)), the library is already used on the site, so no extra page weight
- don't need to initialize `event|volunteering` dates on the modal page, `StartEndCollection` takes care of adding an empty event if needed now

### DIFF
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -172,7 +172,7 @@
           When is it?
         </label>
         <p>You can add multiple dates if the event occurs several times.</p>
-        <StartEndCollection v-if="event.dates" v-model="event.dates" />
+        <StartEndCollection v-if="event.dates" v-model="event.dates" add-date-if-empty />
         <label for="contactname">
           Contact name:
         </label>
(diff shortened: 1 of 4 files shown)


edwh:
Interesting about the cloning.  There are other examples of 
JSON.parse(JSON.stringify) in the code - please can you fix those up 
too?  We should probably always be using cloneDeep.

I checked out your branch and clicked to add a community event, but I 
don't see an empty date slot, so I don't think that's working.  Can you 
check?  For volunteer ops we don't want that function so it is correct 
to lose it.

nicksellen:
Aha, I've fixed it so it doesn't create an empty date for volunteering now.

It works for me, when adding a community event (in chrome and firefox). Maybe the updated code wasn't actually active?

edwh:
Maybe - it certainly works now.  I'll merge.

PR #54, opened 2019-11-15
Add experimental docpages idea

nicksellen:
Just an idea for now...

docpages

### DIFF
--- a/docpages/index.vue
+++ b/docpages/index.vue
@@ -1,0 +1,150 @@
+<template>
+  <div class="w-100">
+    <div class="mx-auto p-5" :style="innerStyle">
+      <h1>Developer docs</h1>
+      <p>The idea is we can use this to write <em>live</em> documentation.</p>
+
+      <b-button v-b-toggle.collapse-group-select block size="lg" class="my-3">
+        GroupSelect
+      </b-button>
+      <b-collapse id="collapse-group-select" class="my-3">
+        <h3>Default mode</h3>
+        <div class="my-3">
+          <group-select v-model="groupSelect.default.value" />
+          <span class="pl-3">value is <code>{{ debugValue(groupSelect.default.value) }}</code></span>
+        </div>
+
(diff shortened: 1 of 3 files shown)


edwh:
I've delayed commenting so I could have a think about it.

This kind of externally facing web-based stuff works well for APIs where you explicitly want people not to have to know about the code internals.  I'm not as keen on it for the code internals themselves.  I tend to look first at the comments when I'm coding, so I can't see it feeling very useful to me personally.  However...I've worked on this project for a long time as a one man band, and so my gut feel may be just plain wrong.  

This example is manually generated, so it becomes another thing to maintain, which would feel like drag.  It's like you said about Swagger, if you have to maintain it manually you're doing it wrong.  But if it was auto-generated from JSDoc comments or similar, so that you were naturally nudged to maintain it as part of maintaining the code, then I'd be keener - it would then improve the quality of the code as well as giving something visual for (relatively) free.

nicksellen:
There are options like https://vue-styleguidist.github.io/ and https://storybook.js.org/ - which can have some elements of auto-generation, but for live examples you tend to need to write some code, it can't really know what kind of data/interaction they should have.

I quite like both of those projects, only having used storybook before, but I fear that the iznik codebase would be too coupled with various store data to get them to work easily. In any case, I don't think they would be _less_ manual work than this approach.

I think this solution doesn't exactly solve my problem/needs either - I think what I want is automated tests, to avoid clicking through 4 different scenarios on the site to see if the `GroupSelect` still works, for example. This solution partially addressed that need as most of the scenarios are here on one page, but it's still not testing.

Given I think it doesn't solve anybodies needs or garner sufficient enthusiasm, I'll close this and we can consider the options in the future.

edwh:
I think that's sensible.  Good mockup though!

nicksellen:
It's not a mockup, it works!

edwh:
:-)

PR #55, opened 2019-11-16
Improved Home Page design for smaller screens

sohcah:
I've improved the Home Page design to work better on smaller screens and updated the Google Play badge.


edwh:
Thanks, I'm away over the weekend but I'll have a look when I get back.

edwh:
I found a little time to have a quick look.

The changes look fine, but we're trying to avoid putting styles directly 
on elements.  Can you move those into a class please, then I'll merge?

The app badges are (slightly oddly) more prominent on desktop than on 
mobile, because on smaller screens they are below the fold.  But I 
intend to do that thing where you prompt people to install the mobile 
app, so that's probably ok.

sohcah:
I'll move the styles to classes in a few hours. 🙂

sohcah:
OK. I've moved the padding-right and padding-left styles into classes.

PR #56, opened 2019-11-16
Fix Login Modal

sohcah:
The Login Modal would not allow you to open it again after being closed. 

The computed `showModal` now has a setter and a getter to fix this issue.


sohcah:
The `index.vue` change was just Nuxt formatting the file when I ran the development server. Forgot to run the development server on my previous commit to make it format the code.

edwh:
Thanks.  The prevention of escape and close is useful too.

PR #57, opened 2019-11-18
Need to tell Vue this is an expression rather than a string

JayyajGH:
This was throwing a warning as it was being passed as a string without the v-bind

### DIFF
--- a/components/GroupHeader.vue
+++ b/components/GroupHeader.vue
@@ -68,6 +68,7 @@
 // TODO DESIGN There's a double scrollbar, e.g. on /explore/ascot-freegle
 // TODO DESIGN There's a max-height hack above to keep the description from hogging the screen.  It's not that
 // pretty and some people may not even notice the scrollbar.  Would be nice to improve it.
+// TODO Add support for showJoin prop
 export default {
   props: {
     id: {
--- a/pages/chats/_id.vue
+++ b/pages/chats/_id.vue
@@ -25,7 +25,7 @@
       <ChatPane v-if="activeChat" v-bind="activeChat" />
     </b-col>
     <b-col cols="0" md="3" class="d-none d-md-block">
-      <SidebarRight :show-volunteer-opportunities="false" show-job-opportunities="true" />
+      <SidebarRight :show-volunteer-opportunities="false" :show-job-opportunities="true" />
     </b-col>
   </b-row>
 </template>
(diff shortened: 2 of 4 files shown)


edwh:
I consistently make this mistake - I wish there was a compile error for 
it.  Could we do anything with eslint or something?

If you search for "false" you'll find some other examples of the same 
bug, e.g.

<groupHeader v-if="group" :key="'groupheader-' + group.id" v-bind="group" show-join="false"/>

Would you be kind enough to review the others and fix them?

JayyajGH:
It is so easy to make!  I'm guessing if an eslint setting existed then it would probably be the default but I'll take a look.  The other option is just not to pass anything in which implies true. e.g. <component showJobOp>.  But obviously that doesn't work for false.
I'll see if I can find/fix any others...

JayyajGH:
In pages/give/whatnext.vue there is this line:
      <groupHeader v-if="group" :key="'groupheader-' + group.id" v-bind="group" show-join="false" />
But showJoin isn't a prop on group header.  Is this an issue or work to do?

nicksellen:
I don't think eslint can do any checks across files, so can't know the types of the components defined in other files.

There is a rule to enforce putting each prop on a new line (that we use in karrot) - https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/max-attributes-per-line.md - this makes it easier to see the props at a glance and in the example in this PR, it would be clearer to see the missing `:`, also helps keep clean diffs (as changing props mean just a line is changed).

edwh:
I'd hope you could have a check which spotted "false" in double-quotes 
as a value to a prop which didn't have a : in front of it.  Regex stuff.

edwh:
I think that's work to do, but which is missing.  Good spot.  Can you 
add a TODO into GroupHeader?

JayyajGH:
Cool will do.  I can't actually find any others with this issue. There are a few going into external components but they would need to investigated individually I think as they might be correct.

JayyajGH:
OK that should now be complete

edwh:
It might just be that one, then.  Thanks for checking.

With "true" it has the same effect, of course.

JayyajGH:
Yeah I checked true as well.

edwh:
Did you fix the invalid settings of the prop in the uses of GroupHeader, 
too?

JayyajGH:
Doh!  Talked about them.  In my head I was going to remove them.  Didn't.  One sec...

JayyajGH:
Yes.  Yes I did. :-)

PR #58, opened 2019-11-18
Use the notice component instead of the b-alert

JayyajGH:
This PR is basically to replace the b-alert with a notice-message component.

The reason this was tricky was because the edit dropdown on the card was floated to the right but the alert component had a higher z index so was shown over it.  The notice component didn't and so the dropdown displayed over it.

Basically I've refactored the news section to use dynamic components and now it only displays the dropdown if the news type is known.

Thoughts, suggestions and abuse please.... :-)

### DIFF
--- a/components/NewsThread.vue
+++ b/components/NewsThread.vue
@@ -3,35 +3,36 @@
     <b-card :style="'background-color:' + backgroundColor" no-body>
       <b-card-body class="p-1 p-sm-2">
         <b-card-text>
-          <b-dropdown class="float-right" right variant="white">
-            <template slot="button-content" />
-            <b-dropdown-item :href="'/chitchat/' + newsfeed.id" target="_blank">
-              Open in new window
-            </b-dropdown-item>
-            <b-dropdown-item :b-v-modal="'newsEdit' + newsfeed.id" @click="show">
-              Edit
-            </b-dropdown-item>
-            <b-dropdown-item @click="unfollow">
-              Unfollow this thread
-            </b-dropdown-item>
-            <b-dropdown-item @click="report">
-              Report this thread or one of its replies
-            </b-dropdown-item>
-            <b-dropdown-item v-if="parseInt(me.id) === parseInt(newsfeed.userid) || mod" @click="deleteIt">
-              Delete this thread
-            </b-dropdown-item>
-          </b-dropdown>
-          <news-message v-if="newsfeed.type === 'Message'" :id="newsfeed.id" :newsfeed="newsfeed" :users="users" @focus-comment="focusComment" />
-          <news-about-me v-else-if="newsfeed.type === 'AboutMe'" :id="newsfeed.id" :newsfeed="newsfeed" :users="users" @focus-comment="focusComment" />
-          <news-community-event v-else-if="newsfeed.type === 'CommunityEvent'" :id="newsfeed.id" :newsfeed="newsfeed" :users="users" @focus-comment="focusComment" />
-          <news-volunteer-opportunity v-else-if="newsfeed.type === 'VolunteerOpportunity'" :id="newsfeed.id" :newsfeed="newsfeed" :users="users" @focus-comment="focusComment" />
-          <news-story v-else-if="newsfeed.type === 'Story'" :id="newsfeed.id" :newsfeed="newsfeed" :users="users" @focus-comment="focusComment" />
-          <news-alert v-else-if="newsfeed.type === 'Alert'" :id="newsfeed.id" :newsfeed="newsfeed" :users="users" @focus-comment="focusComment" />
-          <news-noticeboard v-else-if="newsfeed.type === 'Noticeboard'" :id="newsfeed.id" :newsfeed="newsfeed" :users="users" @focus-comment="focusComment" />
-          <!-- TODO Design - Replace this with a notice-message component. Fix the issue with the floated dropdown above first -->
-          <b-alert v-else variant="danger" show>
+          <div v-if="isNewsComponent">
+            <b-dropdown class="float-right" right variant="white">
+              <template slot="button-content" />
+              <b-dropdown-item :href="'/chitchat/' + newsfeed.id" target="_blank">
+                Open in new window
+              </b-dropdown-item>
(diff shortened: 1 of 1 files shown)


edwh:
I'm a bit surprised we couldn't fix that by setting different z-index values, but you must have considered that.

The alert is really there to spot a coding error while developing the site, and could be dropped once I'm confident I've done them all.  So it depends a bit on whether dynamic component names are a good thing or a bad thing?  I doubt that they'll be as easy to find in an IDE, for example, and it forces us to pass the same props to each.

@nicksellen , do you have a view on that?

JayyajGH:
Yeah we could have fixed it with adding a z-index but I assumed that we didn't actually want the dropdown menu to appear when the news type wasn't recognised.  It would be wrong to use z-index just to hide the dropdown behind the message and not great for accessibility.

My main change was then to only display the components and dropdown for recognised news types.  In doing that I created a function to check what was valid.  This logic was then duplicated.  The function had a list of valid types and the html had the actual list of valid types.  So there was a duplication and a disconnect so they could potentially become out of sync.  Now they are both driven off the same data.

At the moment all of those components have the same props.  If that wasn't the case then this wouldn't be so clean.

I thought the dynamic components might prompt a discussion. :-)

nicksellen:
I think the dynamic component is a big improvement of the tangle of ifs.

I would put it in it's own component though, and consider using a render function, see https://vuejs.org/v2/guide/render-function.html.

Inside it you can do something like:
```js

// import all the news component variants first ...

const types = {
        AboutMe: NewsAboutMe,
        Message: NewMessage,
        CommunityEvent: NewsCommunityEvent,
        VolunteerOpportunity: NewsVolunteerOpportunity,
        Story: NewsStory,
        Alert: NewsAlert,
        NoticeBoard: NewsNoticeboard
}

export default {
  functional: true,
  render (createElement, { data, children }) {
    return createElement(types[data.props.variant], data, children)
  }
}
```

edwh:
Ok, let's do that.

There may be other cases of this, for example in chat.

JayyajGH:
Yeah moving it into a component might be nicer.  I usually prefer the HTML of all my pages/components to fit on one page.

What do we gain from using a render function? I think I'm still from the non-React world where I still prefer using HTML over Javascript to create elements.  I'm not against it.  I'm just trying to understand the advantage.  I've only ever used render functions to create elements where I didn't know the type upfront.

The other thing of interest is that the component also needs to contain the dropdown and notice message as it is all driven off the same data.  That might make using a render function a bit messy.

nicksellen:
> What do we gain from using a render function?

To me, if there is some programmatic logic for which component/element it should be, then a render function is a nice idea. To me, it seems a  bit pointless having a template which only would contain a dynamic `<component :is=...` element, not really much of a template! I would not object to that though, also seems fine :)

> The other thing of interest is that the component also needs to contain the dropdown and notice message as it is all driven off the same data. That might make using a render function a bit messy.

Yes, I think in that case, I wouldn't use a render function, and `<component :is=...` is a good idea when there is other stuff in the template. Also, consider slots if you need other things inside it that can be defined in the parent template, if you don't know about them, worth understanding them, very powerful!

JayyajGH:
Thanks Nick.  Yeah I can see your point.  I'm not particularly bothered either way in the case you mention.  I'll put the logic into a component and lets take a look.  I did consider slots first here but each component was so different it didn't seem that suitable.

I've got quite a lot on over the next week so I'll be dipping in and out of freegle.  I should be able to get to this tomorrow though.

edwh:
Thanks.  I think packing does kinda take  priority.

JayyajGH:
Gonna be without internet for a while (again) too. :-(

edwh:
After I last moved, it took me four years of a community project 
persuading farmers to let us run fibre optic cables across their fields 
before I got decent broadband.

1Gbps symmetric for £30/month now, though, so it was worth it.

JayyajGH:
@edwh  - What is the purpose of the button-content slot template in NewsThread.vue?  Is it just a cut-and-paste thing and can be removed or am I missing something?

edwh:
From memory it was adding a space into the button and I added that in so 
that it would just have the dropdown arrow and align more prettily.

I could be wrong about that memory.

JayyajGH:
Ah ok.  I'll take a look when I'm somewhere I can test the changes.

JayyajGH:
I've been having a quick look at refactoring this PR this morning.  It actually appears that the majority of the NewsThread component is related to this change.  I've tried pulling out the related stuff into a separate component and it actually takes the majority of the original file with it.  Most of the methods etc. are called from this area.

Here's my suggestion...
1. We merge this PR as it is.  It gets rid of a TODO and makes things tidier.
2. I try and refactor the NewsThread component to take out the "Leave a comment" stuff.  This should leave the code related to this PR in this file.  That feels like the best refactor to start with and we can progressively improve it.

Thoughts?

edwh:
1 sounds reasonable.

On 2, have you picked up on the difference between reply to a thread and 
reply to a reply?

JayyajGH:
To be honest I haven't looked into it that far yet.  Is that all dealt with in the card footer?

edwh:
It's inside the card footer in NewsThread, and then there is similar but 
different code inside NewsReply to deal with the reply to reply stuff.

It would be possible to extract the reply function out into a separate 
component, but I think if we did then the component ought to also 
support the similar but different reply function from NewsReply.

That it doesn't feel urgent to me, and some of it is a bit subtle, so 
I'm not sure it's a good thing to do just at the moment.

JayyajGH:
OK.  Maybe I'll just add a TODO to that effect then.  I'll do that now.

JayyajGH:
Could we just take the whole footer out into a separate component?  That would simply both areas nicely and would be a quick first step.

edwh:
I'd rather not.  Then we'd have a situation where the reply was inside 
the NewsReply component but not inside the NewsThread component.  I 
don't find NewsThread too large.  Bung a TODO MINOR in then we can get 
this one merged.

JayyajGH:
Yeah that's fair enough.  TODO added.

edwh:
Ok for me to look at merging?

JayyajGH:
Yup.  Good to go...

PR #61, opened 2019-11-19
Remove the nuxt loading bar

JayyajGH:
Remove the nuxt loading bar as per the TODO.

I've removed the bar in this PR but you can also style it or create your own.
[(https://nuxtjs.org/api/configuration-loading/)]

I'm not sure that it's particularly standard to have a page loading bar.  The browser will tell you the page is loading.  If the page is loading slowly enough to require telling the user then the page load speed should probably be increased. i.e. the task should be to investigate how to make the page load faster (or have a faster perceived load time e.g. image placeholders, load above the fold stuff first etc.)

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -249,7 +249,6 @@
 
 /*TODO DESIGN Menu dropdown on mobile needs a bit of love.*/
 /*TODO DESIGN Shrink navbar on scroll? */
-/*TODO DESIGN MINOR There's a bar that appears at the top when an Ajax request is outstanding.  Probably a Nuxt thing.  It's a nice idea, but irksome.  Maybe just lose it, or have something less annoying on the navbar for the same purpose. /*
  */
 html {
   font-family: 'Source Sans Pro', -apple-system, BlinkMacSystemFont, 'Segoe UI',
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -33,7 +33,7 @@ module.exports = {
   /*
   ** Customize the progress-bar color
   */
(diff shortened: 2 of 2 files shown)


edwh:
I agree with your "speed it up instead" approach in general, but there have been times where it has been very useful to have some kind of indication that an AJAX request is outstanding:

- If we have a spike of load due to TV coverage, then the servers may suddenly become a lot slower than normal.  Things that we would never have felt the need to put any kind of progress indicator into the page for might now be noticeably slow.  
- We have had (and still have) some bugs where the code gets confused and ends up having a flood of requests, or just enough requests to hit the per-host connection limit in the browser, and then things grind to a halt.  HTTP2 helps a bit.  A screenshot from a user can often point us in the direction of there being a bug like that, and we can establish via the user that the loading indicator is stuck on.  That's helped in diagnosing some bugs.
- Where the issue is not our fault, but a slow or flaky network, it helps point the finger.

So this has proved useful in practice.  I would like (but it's TODO MINOR so it can wait) to have a subtle indicator, e.g. a rotating icon on the navbar of a pretty similar colour to the background.  The problem with the default nuxt one is that it was too intrusive, especially on the landing page.

I'll merge this one, but that's the rationale.

JayyajGH:
Sorry, yeah that comment was a bit flippant!  Just speed it up. :-)

I've never done it but it looks like you can create whatever you want fairly easily and hook into the nuxt events.  But that needs thought.  And design.  And space.

edwh:
And time.  Maybe we should add a progress bar to this issue...

PR #62, opened 2019-11-20
Convert another piece of functionality to use the notice component

JayyajGH:


edwh:
Thanks.  What's the click.native change for, so that I understand it?

JayyajGH:
Responding to a native event on a "normal" element like a div you just use @click.  A custom component event handler will only listen to custom events from the component by default.  .native says listen to native events as well.

JayyajGH:
[](https://vuejs.org/v2/guide/components-custom-events.html)

edwh:
Thanks, that's good to know. Will merge.

PR #64, opened 2019-11-20
Remove inline styles

JayyajGH:
Remove the inline style in the component.  More and more colours are coming out of the woodwork!

I've also moved the style section to the bottom of the file as recommended in the Vue style guide.
https://vuejs.org/v2/style-guide/#Single-file-component-top-level-element-order-recommended


edwh:
Ta. All the styles will be in the wrong place, I think.

PR #65, opened 2019-11-20
Add API layer

nicksellen:
For discussion :)

Adds an API layer, to clean up lots of axios request configuration options in the vuex actions.

It implements just `/session` and `/communityevent` API endpoints.

It uses es6 classes, which I would normally avoid, _but_ I wanted a good developer experience, which means autocomplete in editors working. This is actually tricky using nuxt plugins, as you end up making it available as `this.$api` ... and the typing information is not clear.

The _normal_ way to me would be to implement in such a way that you can do `import api from '@/api'` - _but_ axios is only available through the nuxt system, hence needing to implement the API this way too.

In general I'm not such a big fan of es6 classes, in this case, for this reason, maybe it's nice. My first implementation was just as plain objects, but then my IDE just shows me a wiggly line under `this.$api` and offers no completion options.

The mysterious `index.d.ts` enables intellij to have this type information, but I kind of hacked that file together quite ignorantly. I don't know if that would work in other editors too (e.g. vscode).

So, I put this up for initial discussion.

### DIFF
--- a/api/ActivityAPI.js
+++ b/api/ActivityAPI.js
@@ -1,0 +1,7 @@
+import BaseAPI from '@/api/BaseAPI'
+
+export default class ActivityAPI extends BaseAPI {
+  fetch() {
+    return this.$get()
+  }
+}
--- a/api/ActivityAPI.js
+++ b/api/ActivityAPI.js
@@ -2,6 +2,6 @@ import BaseAPI from '@/api/BaseAPI'
 
 export default class ActivityAPI extends BaseAPI {
   fetch() {
-    return this.$get()
+    return this.$get('/activity')
   }
 }
--- a/api/AddressAPI.js
+++ b/api/AddressAPI.js
@@ -1,0 +1,19 @@
+import BaseAPI from '@/api/BaseAPI'
+
+export default class AddressAPI extends BaseAPI {
+  fetch(params) {
+    return this.$get(params)
+  }
+
+  add(data) {
+    return this.$put(data)
+  }
+
+  update(data) {
+    return this.$patch(data)
+  }
+
+  del(id) {
+    return this.$del({ id })
+  }
+}
--- a/api/AddressAPI.js
+++ b/api/AddressAPI.js
@@ -2,18 +2,18 @@ import BaseAPI from '@/api/BaseAPI'
 
 export default class AddressAPI extends BaseAPI {
   fetch(params) {
-    return this.$get(params)
+    return this.$get('/address', params)
   }
 
   add(data) {
-    return this.$put(data)
+    return this.$put('/address', data)
   }
 
   update(data) {
-    return this.$patch(data)
+    return this.$patch('/address', data)
   }
 
   del(id) {
-    return this.$del({ id })
+    return this.$del('/address', { id })
   }
 }
--- a/api/AuthorityAPI.js
+++ b/api/AuthorityAPI.js
@@ -1,0 +1,7 @@
+import BaseAPI from '@/api/BaseAPI'
+
+export default class AuthorityAPI extends BaseAPI {
+  fetch(params) {
+    return this.$get('/authority', params)
+  }
+}
--- a/api/BaseAPI.js
+++ b/api/BaseAPI.js
@@ -1,3 +1,10 @@
+export class APIError extends Error {
+  constructor({ request, response }, message) {
+    super(message)
+    Object.assign(this, { request, response })
+  }
+}
+
 export default class BaseAPI {
   constructor({ $axios, path }) {
     this.$axios = $axios
@@ -5,17 +12,38 @@ export default class BaseAPI {
   }
 
   async $request(method, config) {
-    const res = await this.$axios.request({
+    const { status, data } = await this.$axios.request({
       ...config,
       method,
       url: this.$path,
       baseURL: process.env.API
     })
-    if (res.status !== 200 || res.data.ret !== 0) {
-      console.error('API error', res)
-      throw new Error('API error')
+    if (status !== 200 || data.ret !== 0) {
+      const message = [
+        'API Error',
+        method,
+        this.$path,
+        '->',
+        `ret: ${data.ret} status: ${data.status || 'Unknown'}`
+      ].join(' ')
+      throw new APIError(
+        {
+          request: {
+            path: this.$path,
+            method: method,
+            headers: config.headers,
+            params: config.params,
+            data: config.data
+          },
+          response: {
+            status,
+            data
+          }
+        },
+        message
+      )
     }
-    return res.data
+    return data
   }
 
   $get(params = {}) {
--- a/api/BaseAPI.js
+++ b/api/BaseAPI.js
@@ -6,31 +6,30 @@ export class APIError extends Error {
 }
 
 export default class BaseAPI {
-  constructor({ $axios, path }) {
+  constructor({ $axios }) {
     this.$axios = $axios
-    this.$path = path
   }
 
-  async $request(method, config) {
+  async $request(method, path, config) {
     const { status, data } = await this.$axios.request({
       ...config,
       method,
-      url: this.$path,
+      url: path,
       baseURL: process.env.API
     })
     if (status !== 200 || data.ret !== 0) {
       const message = [
         'API Error',
         method,
-        this.$path,
+        path,
         '->',
         `ret: ${data.ret} status: ${data.status || 'Unknown'}`
       ].join(' ')
       throw new APIError(
         {
           request: {
-            path: this.$path,
-            method: method,
+            path,
+            method,
             headers: config.headers,
             params: config.params,
             data: config.data
@@ -46,16 +45,16 @@ export default class BaseAPI {
     return data
   }
 
-  $get(params = {}) {
-    return this.$request('GET', { params })
+  $get(path, params = {}) {
+    return this.$request('GET', path, { params })
   }
 
-  $post(data) {
-    return this.$request('POST', { data })
+  $post(path, data) {
+    return this.$request('POST', path, { data })
   }
 
-  $postOverride(overrideMethod, data) {
-    return this.$request('POST', {
+  $postOverride(overrideMethod, path, data) {
+    return this.$request('POST', path, {
       data,
       headers: {
         'X-HTTP-Method-Override': overrideMethod
@@ -63,19 +62,19 @@ export default class BaseAPI {
     })
   }
 
-  $put(data) {
-    return this.$postOverride('PUT', data)
+  $put(path, data) {
+    return this.$postOverride('PUT', path, data)
   }
 
-  $realPut(data) {
-    return this.$request('PUT', { data })
+  $realPut(path, data) {
+    return this.$request('PUT', path, { data })
   }
 
-  $patch(data) {
-    return this.$postOverride('PATCH', data)
+  $patch(path, data) {
+    return this.$postOverride('PATCH', path, data)
   }
 
-  $del(data, config = {}) {
-    return this.$postOverride('DELETE', data)
+  $del(path, data, config = {}) {
+    return this.$postOverride('DELETE', path, data)
   }
 }
--- a/api/BaseAPI.js
+++ b/api/BaseAPI.js
@@ -39,6 +39,10 @@ export default class BaseAPI {
     return this.$postOverride('PUT', data)
   }
 
+  $realPut(data) {
+    return this.$request('PUT', { data })
+  }
+
   $patch(data) {
     return this.$postOverride('PATCH', data)
   }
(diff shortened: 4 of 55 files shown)


edwh:
Thanks, I'll have a look today. 

What do you have against ES6 classes?

edwh:
I'm glad we're doing this, as the axios code is verbose and we'll end up with something that's nicer.

I'm not (yet) convinced by the layer of an API for a single call/area.  There's a risk that this is boilerplate pass-through code which doesn't add much value on top of what the store logic will need to have.   

However this may change because auth/session is an unusual case, whereas a lot of them will be more like the CommunityEvents one.  I suspect that a lot of the calls could be handled by a fairly generic bit of code which you then just extended with the name of the API call and maybe another method or two which are specific to that call.

So could you try adding a couple more calls in there and see how the code evolves?

Nit: the exception should probably include more info.

nicksellen:
I added the /message endpoint - I left all then `.then().catch()` stuff in place, but would really turn that into async/await to make it cleaner.

It would totally be possible to build generic code that is then extendable, a few approaches:
- different base classes (e.g. `SomeOtherAPI extends DefaultAPI`), downside is only supports single inheritance, and I hate inheritance anyway (composition > inheritance), plus side is IDE should still be able to understand which methods are available
- mixins (something like `Object.assign(MyClass.prototype, defaultMethods)`), allows multiple mixins, but probably lose editor support
- normal functions in `api/index.js` ( e.g. `{ message: { ...defaultMethods, ...someOtherMethods } }`)

Having said that, I am not so much in favour of magic method creation like this, as the API classes also are a kind of simple documentation of what is available, and a place to add jsdoc/parameter definitions, etc... so I'm more for explicit methods, but with helpers to make them as short as possible.

I agree they are a bit boilerplate-y (like much of vuex too), I see more the point though, to seperate concerns, and handle various things consistently (e.g. checking response for errors).

I don't like es6 classes in general as I don't like inheritance in general, javascript has plenty of methods already for doing what classes can do, but in this case I would be in favour of them (as they help keep the type information).

nicksellen:
> Nit: the exception should probably include more info.

Yes, I agree! Actually adding sentry would be a good idea (I saw it mentioned elsewhere), and also some UI for users to know something might have gone a bit wrong (e.g. [a toast](https://bootstrap-vue.js.org/docs/components/toast)). At the moment in nuxt dev it'll print generic "API Error" and have full details in console.

edwh:
I agree with using async/await.  I hadn't used that in earnest before 
this rewrite, so early on I didn't.

I'm happy for you to decide which way to go on whether/how to do a 
generic bit of code - just wanted to check you were thinking of it 
(sorry).

edwh:
Yes, I have a todo for Sentry somewhere.

There are various TODOs about error case checking.  I'm leaning more now 
towards:
spotting all API errors and JS exceptionslogging them somewhere so that 
we can see them - I have server API logs, but Sentry is very nice for 
JShaving a generic "something went a bit wrong" toast
One reason for this is that coding for all the error cases all over the 
place is a) dull b) expensive c) hard to test d) unlikely to happen.  So 
a generic "something went wrong" solution is quite appealing.

What do you think about that?

nicksellen:
I made it show a toast for uncaught API exceptions - in some cases you want to catch the error somewhere and do some different logic, so this is handling it at the top level that vue makes available (`Vue.config.errorHandler`). If it's an API error, then show toast, otherwise do whatever it would have done before (nuxt error handling).

The message is not very good, maybe you want to come up with something, the toast has a title and a message.

nicksellen:
This being what I made it do for now:

apierror

JayyajGH:
Probably not going to get a chance to have a look at the code properly but I'm a big fan of abstracting the axios stuff out. I don't think the store should be tied to a particular method of making API calls.  I've done a very crude version of this in the past myself too.

edwh:
Can you check how the toast would show if we were inside a modal?

I was wondering whether the reason that the API layer feels empty is that there is a fair amount of API knowledge still in the pages via the API parameters.  It's the page that knows what parameters to pass into the store to pass into the API to make the call, which kind of makes a mockery of those being separate layers - they're passing through the interesting stuff without being aware of it.

An example is in messages - you use the same call to fetch your own posts, the posts on a group, the posts on all groups, the posts by a specific user (we might not have an example of that).  (Let's put aside any possible discussion about whether that's because the API is badly designed - for the purposes of this rewrite the API is largely fixed except for backwards-compatible enhancements.)

Perhaps the pages should be calling the store (which calls the API) saying what they want to achieve, not passing the specific API parameters.  What do you think about moving that awareness out of the pages and down into the API layers?

nicksellen:
I would be very happy to have more specific API methods instead of generic parameters, actually I think it kind of highlights the use of an API layer, at the moment it's hard for me to know what are valid parameters to pass through.

The `session/login` one was the only one where I could see the parameters, although there may be others. And it could be split into seperate methods too (`this.$api.session.facebookLogin()`, etc...).

These kind of changes aren't too tricky to implement though, if we agree on the _basic_ premise of the API layers, so if you do, I would suggest I proceed to create the remaining API classes and get it merged, and refine/improve after that.

If you do agree to that plan, then other than adding more API classes, what else needs to be done before that? Off the top of my head:
- refine the API error toast message
- check API error toast is reasonable with modal displayed
- check that any places where I've removed the return value in actions, that it's not actually needed, if so provide equivalent (or refactor usage)

edwh:
> I would be very happy to have more specific API methods instead of generic parameters, actually I think it kind of highlights the use of an API layer, at the moment it's hard for me to know what are valid parameters to pass through.

Yes.

> 
> The `session/login` one was the only one where I could see the parameters, although there may be others. And it could be split into seperate methods too (`this.$api.session.facebookLogin()`, etc...).
> 
> These kind of changes aren't too tricky to implement though, if we agree on the _basic_ premise of the API layers, so if you do, I would suggest I proceed to create the remaining API classes and get it merged, and refine/improve after that.
> 

Agree.

> If you do agree to that plan, then other than adding more API classes, what else needs to be done before that? Off the top of my head:
> 
> * refine the API error toast message

"Sorry, something went wrong.

That might be a bug.  Please try again - if you continue to have problems then please take a screenshot and contact support@ilovefreegle.org."

That'll do for now.

> * check API error toast is reasonable with modal displayed

We might need to reposition it

> * check that any places where I've removed the return value in actions, that it's not actually needed, if so provide equivalent (or refactor usage)

There are some cases where I return from an action the object I have refetched into the store.  Is that what you mean?  I think that's probably a good convention even if we don't always use it.

nicksellen:
Great :+1: 

> We might need to reposition it

I would _guess_ a global toast will appear on top of everything. There are 8 default positions where you can put (see https://bootstrap-vue.js.org/docs/components/toast/). It also says it'll have default z-index of 1100.

> There are some cases where I return from an action the object I have refetched into the store. Is that what you mean? I think that's probably a good convention even if we don't always use it.

I kept the return if it's actually an interesting object, but there  are other cases where it returns the axios response object, which we no longer have here. For now, I just removed those returns, but I need to check if they were actually used first. If they were, it would probably be to use an interesting value, not the axios stuff, in which case I can replace it easily.

edwh:
It might appear on top of it, but your eye will be focussed on the 
modal, so if there's a toast at the bottom of the screen not everyone 
will notice it.  We would really want to toast to be near the modal 
footer, I think.

On the return value - yes, that all sounds sensible.

nicksellen:
So, API classes for all the endpoints created - originally I was keeping to one path = one API class, but after seeing the various `/chats*` endpoints that didn't make sense to me any more, so some of them will make requests to different endpoints where seems sensible - to provide a slightly more rational view on top of the backend API.

I updated the toast error message too.

I briefly looked into making it appear near a modal, but I don't think this is worth it for the complexity it would involve - most of the modals I can find are quite big anyway (so it would be at the bottom), and the small ones I could find don't seem to do API requests... I'm sure there are some, but seems marginal benefit for the added complexity, when we shouldn't be triggering these errors anyway (or relying on users to act on them - e.g. by using sentry).

I would do one more details look through to try and catch any errors I have introduced by converting them, but other than that I think this is good to go, but please add any more feedback if you have some :)

edwh:
It'll be a few days before I can review that properly.  Can you refresh 
against the latest codebase if you haven't recently, as you'll pick up 
new stores.

nicksellen:
No problem :) It's all rebased on top of current master, so anything new is in there.

edwh:
Ok, good.  I may be able to take a look tomorrow.

edwh:
I thought it would make sense to have something like BaseAPI.  The 
reason we use the override mechanism for HTTP verbs is that I have seen 
some scenarios, I think in corporate environments, where those HTTP 
verbs were blocked.  Mad, I know.  We should always be using GET/POST - 
we weren't, which is why you added realPut, but that was a bug in the 
existing code so please remove that method and replace with a call to 
the patch version.

I wonder whether you could push BaseAPI further though.  If you look at 
(say) AddressAPI and ActivityAPI, then they are basically the same apart 
from a) the name of the call and b) some additional methods.  It might 
reduce boilerplate code if we moved the name of the API call into Base, 
so that the simpler API calls just extend that and set the API call 
property - if there are additional methods which are not used or even 
not implemented on the server I think that's fine.  Probably you could 
push this even further to avoid a bunch of individual files for each 
one.  But I'll let you decide whether it's worth doing this (which you 
may already have done).

Entirely sensible to combine the chat calls into one API, thanks.

We have index.d.ts and index.js, which are basically the same.  There 
will be a reason for this, which you've probably already told me - but 
put it in a comment in both of them to explain why and ensure that 
people don't just change one of them.

In the toast, I have felt bad about having Freegle-specific URLs and 
email addresses in templates and code.  Initially I felt bad because I 
hoped other people might use the code, but that isn't happening.  If you 
feel bad about this too, what should we do?

I've not pored over all the store changes to use the API.  They look 
fine and the site runs ok from a quick test, so I'd rather get this 
merged and shake out any bugs later.

nicksellen:
I think I would be up for convenience base methods that are _only_ called from the concrete API classes - perhaps they would have the `$` prefix (e.g. `$fetch`), however, I'm not sure these will be that useful. I would like in the future for the public API methods to get more specific about their accepted params, and maybe more methods for specific types of update, etc... 

Although if there really were some very standard sets of API methods, creating another base class (e.g. `StandardAPI extends BaseAPI` might be an option... but I'm also wary of class hierarchies growing too much (or random stuff being thrown into `BaseAPI` through lack of imagination!).

Yeah, I'm not super happy with the ` index.d.ts` at all (it's just for IDE typing information), and also it's very similar content to `index.js` - I think there might be scope for improvement, but for now I agree with your suggestion of a explanatory note - I'll add that.

I don't feel bad about Freegle specific information at all - it would not be hard to extract that to a configuration option, and given there are not others using it, I don't think it's a good idea to implement things very hypothetically. I think it would be a nice idea that others use it, but only making the effort in the code (and not the other non-code/tech things to encourage adoption) doesn't make sense to me. That's my thoughts anyway.

Yup, I'm up for getting it merged sooner than later - it's definitely not perfect, there are some slightly odd calls where I didn't quite understand what the meaning/purpose of the call is, but in those cases it was still possible to feel reasonably comfortable that the functionality will still work.

edwh:
Ok.  So shall I merge now?

nicksellen:
No, not yet, just addressing the duplication of index.js and index.d.ts...

nicksellen:
OK, I created a generator for them :)

edwh:
That generator is nice, thanks.  Is there any way to integrate it into the build so it runs automatically?  You may have done that, but I can't see it.

If not easy, then that's fine - just asking before I merge.

nicksellen:
It won't need to be run very frequently, and I would more consider those files part of the source, rather than a build output. So, I would suggest merging as it is now :)

edwh:
I've merged this.

I briefly saw the toast because my client was logged out, though the server was logged in.  I haven't quite got a repro scenario, but I think we should not through the APIError for data.ret === 1 (which means you're not logged in).  There's separate code to handle that.  So I'll push that change, but shout if you disagree.

nicksellen:
Yes, this is a good scenario where the new error handling can make sense - if in a particular action, an APIError does not mean it's an _actual_ error, you can catch it and handle it differently.

The toast error is just when it's not expected.

e.g. something similar to

```js
myAction() {
  try {
    this.$api.foo.bar()
  } catch (err) {
    if (err instanceof APIError) {
      // don't worry all is fine
    } else {
      throw err // rethrow...
    }
  }
}
```

From the API perspective (which the API error is encapsulating) it's still an error, just not at the application/action level. The goal is the separation of concerns.

Does it make sense?

edwh:
Yes, there are times when it would be fine to catch the error.

I think where there's a global feature of the API (such as "ret === 1 
means not logged in") then we would want to reflect that in the API 
layer.  So in this case, perhaps we should be throwing a "not logged in" 
exception which people could then catch or not.

But that is a larger change because of the interaction with the 
middleware, and I just wanted to get the toast off the screen in this 
common case for now.

PR #66, opened 2019-11-21
Allow disabling of eslint autofix with local conf

nicksellen:
Adds ability to disable eslint autofixing.

If you want eslint autofix do nothing, it will work as before.

If you DO NOT want eslint auto fix, add a `.env` file in the root of the project with the contents:

```
DISABLE_ESLINT_AUTOFIX=true
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /debug.log
 /modules/lazystrap/node_modules/.cache
 /.idea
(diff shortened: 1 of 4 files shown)


edwh:
Will merge.

How will people know to do this, though?  Will you put this in that 
thing you're writing?

PR #71, opened 2019-11-29
Change the anchors to buttons to be semantically correct

JayyajGH:


edwh:
I hadn't noticed variant=link.  That great, thanks.  There are definitely other examples of that kind of thing in the codebase, e.g. in LoginModal.  I'm not sure what the href="#" is doing in WizardProgress either.  I'll merge this one.

PR #72, opened 2019-11-29
Remove any duplicates from the notification list

JayyajGH:
This PR deals with the TODO in default.vue regarding the Vue warning about duplicates in the notifications list.

The duplicates in default.vue notifications were coming from the store.  Each time the 'list' store action was called it was calling the 'add' mutation.  The lodash 'union' method combined the old and new notification data together.  Union only removes duplicate on primitive values and so the duplicate objects weren't being removed.  I've added an comparator function in to compare the IDs.

You might want to take a proper look at this solution as I'm not that familiar with the way the store is being used with the axios calls and when the data is refreshed. e.g. why don't we clear the list every time?

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -68,7 +68,6 @@
               </b-dropdown-item>
               <b-dropdown-divider />
               <b-dropdown-item v-for="notification in notifications" :key="'notification-' + notification.id" class="p-0 notpad">
-                <!--                TODO We're getting duplicate key errors here - click on notifications, close, click again.  Why?-->
                 <Notification :notification="notification" class="p-0" @showModal="showAboutMe" />
               </b-dropdown-item>
               <infinite-loading :distance="distance" @infinite="loadMore">
--- a/store/notifications.js
+++ b/store/notifications.js
@@ -1,5 +1,6 @@
 import Vue from 'vue'
 import union from 'lodash/union'
+import uniqWith from 'lodash/uniqWith'
 import twem from '~/assets/js/twem'
 
 export const state = () => ({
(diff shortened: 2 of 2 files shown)


JayyajGH:
Another related question...

What is the purpose of the "object.values" in the 'Notifications" method in default.vue?  The store getter just returns an array anyway doesn't it?

edwh:
From my phone so written without looking at the code. 

I think we ought to be clearing each time, when the event to show the drop down is triggered. Generally the pages take the approach of clearing out the store on load or significant change. 

More generally it would be better to leave the entries in the store for speed of render, and then they would update with new/changed info when the server API call completes. That would need all stores to merge rather than add, but also have a way to spot entries which need removing because they are no longer returned from the server, which would be a bit complex.

So in this case it's probably better to remove your current fix, add a clear action if there isn't one, and then invoke that on the dropdown event.

The Object.values is probably a copy/paste error. In stores which are persisted we use objects rather than arrays (to avoid sparse arrays which serialise the badly). In any areas to do with scanning arrays/properties/enumerating, if it looks like I've done something stupid then I probably have.

JayyajGH:
Thanks for your response.  I'll read and digest this properly on Monday.  It was very specifically coded with a union and so it appeared to be deliberate.  Maybe take a look at the code when you have the chance to confirm that clearing would be the right option and I will implement next week.

edwh:
Yeah, no rush. I'll have a look and get back to you.

I suspect the union was a half baked attempt to render rapidly by preserving the old contents. It might be OK because notifications (unlike other items) can never be deleted. In which case your fix as it stands might do the trick nicely for now.

edwh:
I've merged this and added a comment.

JayyajGH:
Thanks for adding the comment.  Your notes and comment all make complete sense.

If you are using an array here rather than an object then should we also get rid of the "object.values" code in default.vue?

edwh:
Ah, I forgot about that. Yes, I'll do that.

JayyajGH:
Cool.   Thanks Edward!

PR #73, opened 2019-12-02
Remove the footer duplication and move into a component

JayyajGH:
The footer in index had the options duplicated - one for desktop and one for mobile.  I've consolidated these into a single list which display differently at different breakpoints.

NB: I'm using the bootstrap functions to get the breakpoints rather than hardcoding the values in here.


edwh:
Thanks.  I like the use of bootstrap breakpoints - there's a TODO for that in global.css and NewsReply which you could apply the same logic to.

PR #74, opened 2019-12-02
Use bootstrap breakpoints

JayyajGH:
Use the bootstrap breakpoints where necessary.

I've added comments for other places to do this i.e. the profile classes.  I might break the profile image stuff out into a component next so I'll sort the breakpoints for those classes then.

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -143,6 +143,7 @@ body {
   width: 35px !important;
   height: 35px !important;
 
+  // TODO design Use bootstrap breakpoints
   @media (min-width: $bootstrap-sm) {
     width: 50px !important;
     height: 50px !important;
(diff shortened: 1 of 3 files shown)


edwh:
Can you double-check these are the right way round?

In NewsThread @media (max-width: 992px) becomes @include 
media-breakpoint-up(lg).  My reading of 
https://getbootstrap.com/docs/4.3/layout/overview/ is that the -up means 
"from this breakpoint upwards", whereas max-width means downwards.

JayyajGH:
I tested that as well but still got it the wrong way around!  NewsThread should now be fixed.

JayyajGH:
I re-checked NewsReply and it looks ok.  Although the leaf isn't in the right place in the smallest breakpoints but that's another issue.

edwh:
Hmmm, I thought I'd fixed that recently.  Will refix.

PR #75, opened 2019-12-02
A few api changes

nicksellen:
For the JobAPI:
- need to pass in params in as { params } to axios calls
- get missing `status` value to use in case of error
- add extra data in APIError to match the BaseAPI style
- remove redundant try/catch in action
- don't inherit from BaseAPI (as it doesn't use anything from there)

.. I also removed the `$realPut` so it _always_ uses method overide approach now.

### DIFF
--- a/api/BaseAPI.js
+++ b/api/BaseAPI.js
@@ -69,10 +69,6 @@ export default class BaseAPI {
     return this.$postOverride('PUT', path, data)
   }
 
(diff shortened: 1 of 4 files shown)


nicksellen:
Ah, we duplicated on removing real put :) https://github.com/Freegle/iznik-nuxt/commit/ded5b05a6e09a4d7000bd864d67ce0706c509ca1

PR #76, opened 2019-12-02
Add form validation

nicksellen:
So, the idea for this is to use vuelidate, I'm already familiar with it, and it's the first suggestion given on the [boostrap-vue validation page](https://bootstrap-vue.js.org/docs/reference/validation/#form-validation) too.

One of the things that it doesn't do for you is the error message handling, leaving that to you - which is a reasonable idea as there are many ways to show/display/fetch the actual error messages. The downside is it can get quite verbose and repetitive. So, I thought to create new form components that wrap the bootstrap ones to take the tedium out of it.

It basically passes everything through to the underlying component, apart from the props `validation` (the vuelidate object representing a specific field) and `validation-messages` (a map of validation type -> message) which it uses itself.

I used the same pattern as shown in the bootstrap-vue docs, so it looks a bit like this:

validation

How do you feel about this general approach (wrapping the bootstrap form components)?

The other question that quickly arises is what the validation rules actually be? I wonder if there is a reference somewhere I can follow? (maybe the backend has some hints? or the old frontend?).

### DIFF
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -39,6 +39,7 @@ $color-yellow-2: #FFEEBA;
 $color-yellow-3: #856404;
 
 $color-red: #FF0000;
+$color-danger: #dc3545; // from bootstrap _variables
 
 $color-white: #FFF;
 $color-black: #000;
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -39,7 +39,6 @@ $color-yellow-2: #FFEEBA;
 $color-yellow-3: #856404;
 
 $color-red: #FF0000;
-$color-danger: #dc3545; // from bootstrap _variables
 
 $color-white: #FFF;
 $color-black: #000;
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -100,6 +100,7 @@
         <b-row>
           <b-col cols="12" md="6">
             <b-form-group
+              ref="groupid"
               label="For which community?"
               :state="validationEnabled ? !$v.groupid.$invalid : null"
             >
@@ -110,6 +111,7 @@
             </b-form-group>
             <b-form-group
               v-if="enabled"
+              ref="eventEdit__title"
               label="What's the event's name?"
               label-for="title"
               :state="validationEnabled ? !$v.eventEdit.title.$invalid : null"
@@ -175,6 +177,7 @@
             </b-col>
           </b-row>
           <b-form-group
+            ref="eventEdit__description"
             label="What is it?"
             label-for="description"
             :state="validationEnabled ? !$v.eventEdit.description.$invalid : null"
@@ -195,6 +198,7 @@
             />
           </b-form-group>
           <b-form-group
+            ref="eventEdit__location"
             label="Where is it?"
             label-for="location"
             :state="validationEnabled ? !$v.eventEdit.location.$invalid : null"
@@ -212,6 +216,7 @@
             />
           </b-form-group>
           <b-form-group
+            ref="eventEdit__dates"
             label="When is it?"
             :state="validationEnabled ? !$v.eventEdit.dates.$invalid : null"
           >
@@ -222,6 +227,7 @@
             <StartEndCollection v-if="eventEdit.dates" v-model="eventEdit.dates" add-date-if-empty />
           </b-form-group>
           <b-form-group
+            ref="eventEdit__contactname"
             label="Contact name:"
             label-for="contactname"
             :state="eventEdit.contactname && validationEnabled ? !$v.eventEdit.contactname.$invalid : null"
@@ -357,7 +363,6 @@
 // TODO NS Don't allow submission before image upload complete.
 // TODO Wherever we have b-img (throughout the site, not just here) we should have @brokenImage.  Bet we don't.
 // TODO NS Set date to start at 9am rather than midnight.  Default end date to later than start date.
-import Vue from 'vue'
 import { required, maxLength } from 'vuelidate/lib/validators'
 import cloneDeep from 'lodash.clonedeep'
 import { validationMixin } from 'vuelidate'
@@ -369,15 +374,6 @@ const GroupRememberSelect = () => import('~/components/GroupRememberSelect')
 const OurFilePond = () => import('~/components/OurFilePond')
 const StartEndCollection = () => import('~/components/StartEndCollection')
 const NoticeMessage = () => import('~/components/NoticeMessage')
-
-// TODO NS move into utilities somewhere
-function nextTicks(n, fn) {
-  if (n === 0) {
-    fn()
-  } else {
-    Vue.nextTick(() => nextTicks(n - 1, fn))
-  }
-}
 
 function initialEvent() {
   return {
@@ -484,9 +480,7 @@ export default {
     async saveIt() {
       this.$v.$touch()
       if (this.$v.$anyError) {
-        // It takes a few cycles to actually render the errors...
-        // TODO NS this is a bit hacky, I think a better way is to use refs to find the elements to focus
-        nextTicks(4, () => this.validationFocusFirstError())
+        this.validationFocusFirstError()
         return
       }
 
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -396,8 +396,8 @@ export default {
       this.hide()
     },
     async saveIt() {
-      this.$v.form.$touch()
-      if (this.$v.form.$anyError) {
+      this.$v.event.$touch()
+      if (this.$v.event.$anyError) {
         return
       }
 
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -105,13 +105,17 @@
             <label v-if="enabled" for="title">
               What's the event's name?
             </label>
-            <b-form-input
+            <validating-form-input
               v-if="enabled"
               id="title"
               v-model="event.title"
               type="text"
-              maxlength="80"
               placeholder="Give the event a short title"
+              :validation="$v.event.title"
+              :validation-messages="{
+                required: 'Title is required',
+                maxLength: ({ max }) => `Max length is ${max}`
+              }"
             />
           </b-col>
           <b-col v-if="enabled" cols="12" md="6">
@@ -163,7 +167,7 @@
           <label for="description">
             What is it?
           </label>
-          <b-textarea
+          <validating-textarea
             id="description"
             v-model="event.description"
             rows="5"
@@ -171,11 +175,24 @@
             spellcheck="true"
             placeholder="Let people know what the event is - why they should come, what to expect, and any admission charge or fee (we only approve free or cheap events)."
             class="mt-2"
+            :validation="$v.event.description"
+            :validation-messages="{
+              required: 'Description is required'
+            }"
           />
           <label for="location">
             Where is it?
           </label>
-          <b-form-input id="location" v-model="event.location" type="text" maxlength="80" placeholder="Where is it being held?  Add a postcode to make sure people can find you!" />
+          <validating-form-input
+            id="location"
+            v-model="event.location"
+            type="text"
+            placeholder="Where is it being held? Add a postcode to make sure people can find you!"
+            :validation="$v.event.location"
+            :validation-messages="{
+              required: 'Location is required'
+            }"
+          />
           <label>
             When is it?
           </label>
@@ -271,8 +288,12 @@ label {
 // TODO NS Don't allow submission before image upload complete.
 // TODO Wherever we have b-img (throughout the site, not just here) we should have @brokenImage.  Bet we don't.
 // TODO NS Set date to start at 9am rather than midnight.  Default end date to later than start date.
+import { validationMixin } from 'vuelidate'
+import { required, maxLength } from 'vuelidate/lib/validators'
 import cloneDeep from 'lodash.clonedeep'
 import twem from '~/assets/js/twem'
+import ValidatingFormInput from '@/components/ValidatingFormInput'
+import ValidatingTextarea from '@/components/ValidatingTextarea'
 const GroupRememberSelect = () => import('~/components/GroupRememberSelect')
 const OurFilePond = () => import('~/components/OurFilePond')
 const StartEndCollection = () => import('~/components/StartEndCollection')
@@ -280,11 +301,14 @@ const NoticeMessage = () => import('~/components/NoticeMessage')
 
 export default {
   components: {
+    ValidatingFormInput,
+    ValidatingTextarea,
     GroupRememberSelect,
     OurFilePond,
     StartEndCollection,
     NoticeMessage
   },
+  mixins: [validationMixin],
   props: {
     event: {
       type: Object,
@@ -372,6 +396,11 @@ export default {
       this.hide()
     },
     async saveIt() {
+      this.$v.form.$touch()
+      if (this.$v.form.$anyError) {
+        return
+      }
+
       // TODO NS Validation.
       this.saving = true
 
@@ -501,6 +530,20 @@ export default {
     },
     rotateRight() {
       this.rotate(-90)
+    }
+  },
+  validations: {
+    event: {
+      title: {
+        required,
+        maxLength: maxLength(80)
+      },
+      description: {
+        required
+      },
+      location: {
+        required
+      }
     }
   }
 }
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -312,6 +312,8 @@
 
 <style scoped lang="scss">
 @import 'color-vars';
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
 
 .field {
   font-weight: bold;
@@ -330,7 +332,7 @@
 
 ::v-deep .is-invalid label,
 ::v-deep .is-invalid .col-form-label {
-  color: $color-danger;
+  color: $form-feedback-invalid-color;
 }
 
 ::v-deep .form-group.is-invalid .invalid-feedback {
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -354,7 +354,6 @@
 
 <script>
 // TODO DESIGN This layout is staid table nonsense.  Surely we can make it more appealing?
-// TODO NS Add some form validation using a plugin - see https://bootstrap-vue.js.org/docs/reference/validation/
 // TODO NS Don't allow submission before image upload complete.
 // TODO Wherever we have b-img (throughout the site, not just here) we should have @brokenImage.  Bet we don't.
 // TODO NS Set date to start at 9am rather than midnight.  Default end date to later than start date.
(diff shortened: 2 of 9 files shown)


edwh:
Happy with this general approach, as long as there is some way to result 
in an overall check for the whole form.  It looks like you have that in 
the example for when you click - I don't quite follow how it works, but 
I can pick that up later - but can you also do it dynamically, e.g. to 
enable/disable a button once the inputs become valid/invalid?

The error messages need to be nice and warm, e.g. "Please add a title" 
is nicer than "Title is required".

There's no real documentation of the validation required.  You could 
look in the old code, and there are some notes in some of the todos.  
It's probably easiest to do what seems obvious and I'll review it - once 
you've put the mechanism in place it'll be easy to add rules.

nicksellen:
> Happy with this general approach, as long as there is some way to result in an overall check for the whole form.  It looks like you have that in the example for when you click - I don't quite follow how it works, but I can pick that up later - but can you also do it dynamically, e.g. to enable/disable a button once the inputs become valid/invalid?

Yup, you can really do it any way desired with this stuff, it's all manually triggered - in this case I've hooked into the _blur_ events on the fields to trigger validation on those specific fields (e.g. `$v.event.title.$touch`), and the `saveIt` method triggers all the fields with `$v.event.$touch`). It also keeps track of which fields are _dirty_ (changed).

Although if the button is disabled until everything is valid, it's not clear how to trigger validations for fields which you haven't touched yet, and pre-emptively touching them all makes the whole thing go red and scary, so I think keeping the button enabled might be better.

But it could work with a disabled the button if it's very clear which fields are required, so the user is likely to touch them all before wanting to attempt a save.

Maybe a bit of UX research would come up with a good approach...

> There's no real documentation of the validation required.  You could look in the old code, and there are some notes in some of the todos. It's probably easiest to do what seems obvious and I'll review it - once you've put the mechanism in place it'll be easy to add rules.

Ok, I'll probably keep it minimal in that case, and see about getting this community event form mergable. Then I can proceed to the basic stuff to other forms, and maybe you can add in various rules. I suspect there will be some more fiddly ones though... e.g. `StartEndCollection` - I guess for community event there should be at least 1 date pair, and each should have a start and end? ... I'd also prevent adding in end dates earlier than start dates, shouldn't wait for validation to catch that.

edwh:
There may be a standard way to indicate mandatory/optional stuff nowadays, though I've seen so many variations that I doubt it.

Probably keeping enabled is OK, but Google a bit.

Yes, events have at least one date, end after start, gap between not too long. Thing there may be a TODO about that.

I had some code to change the end date when you set the start date to ensure it was always after. That had issues though and would need doing carefully. That aspect can wait.

JayyajGH:
My first thought was is it getting a bit like a rabbit hole with components wrapping bootstrap components wrapping HTML elements?  It's probably best to ignore that opinion though as I'm not a big fan of bootstrap (have I mentioned that before? :-p)

My preference for validation is to always have the "Continue/Submit" button enabled rather than disabling it until all fields are valid/entered.  There is nothing worse than not being able to see what you have missed.  This tends to create a much more accessible experience too.  This also fits in with how Nick was suggesting it would work technically.

I've got a bunch of docs saved somewhere on the topic of validation/forms.  I'll try and dig some out...

JayyajGH:
Adam silver is a well respected form designer. This link has loads of great info:
https://adamsilver.io/articles/form-design-from-zero-to-hero-all-in-one-blog-post/?ref=heydesigner

Some of that is also backed up here:
https://uxmovement.com/forms/why-users-make-more-errors-with-instant-inline-validation/

edwh:
Interesting stats on instant validation in the second link.

JayyajGH:
This is obviously the internet so everything has to be taken with a pinch of salt.  Adam Silver stuff is usually pretty well backed up and referenced though so I do tend to trust what he says.

JayyajGH:
> I had some code to change the end date when you set the start date to ensure it was always after. That had issues though and would need doing carefully. That aspect can wait.

I think this needs to be thought about carefully.  Changing something automatically that the user has entered could be confusing.  e.g. someone sets up an event for a few weekends time - 21/12 - 22-12.  They then realise they meant the weekend before so want to change the 22nd to 15th.  And then before they can change 21st to 14th  the date they have just changed automatically updates to something else.

edwh:
Yes, it needs to be careful and is a separate PR from this initial 
validation.  I think probably we shouldn't change the end date once they 
have set an end date, but I think it's helpful to change the end date 
when they first change the start date.

It helps a lot with events which are on the same day (which is the vast 
majority) because it avoids having to do the tedious month and date 
selection twice, leaving you to just focus on the time.

JayyajGH:
> but I think it's helpful to change the end date
> when they first change the start date.

Yeah absolutely.  My form hatred is when you set a start date and the end date stays at the current date.  ARGH!

nicksellen:
> My first thought was is it getting a bit like a rabbit hole with components wrapping bootstrap components wrapping HTML elements? It's probably best to ignore that opinion though as I'm not a big fan of bootstrap (have I mentioned that before? :-p)

I think you just described programming... :) It's one big rabbit hole. And it doesn't stop at the HTML element either... if we weren't using bootstrap we'd be using either another UI framework with it's own components, or writing our own reusable components, so it's not much different in any case. 

This approach to wrap components is the standard vue approach: _Note that in Vue 2.0, the primary form of code reuse and abstraction is components_ (on https://vuejs.org/v2/guide/custom-directive.html - as I was wondering whether a directive could do the same thing as this component).

The idea is that each component is doing one thing, and if you are not thinking about that particular aspect at one moment, you don't need to open up that component - so long as you trust it does what it promises to do.

The opposite, putting everything directly in the lower down component means it would be very verbose and a lot of duplication of code. Maybe thinking of components like functions helps? (if you had the same code in a few places, you'd abstract it into a function, which is also a "rabbit hole" of functions, I like rabbit holes anyway, I think Alice had a much more interesting time after choosing that option).

Also, thanks for form thoughts, that's exactly the kind of thinking I was hoping you might have :)

I'm wondering how to make it concrete - to implement Adam Silvers ideas we'd have to implement a lot of our own components, which we could do, but it would be quite a lot of work. On the other end of the scale we could just hold off validation until the user first presses the submit button (the suggestion from the other article), but keep using all the bootstrap components as already exist.

Thoughts?

JayyajGH:
Ha!  Yeah good point!  I think as you say, the main thing is to keep things DRY.  And if that means creating more components then that is what is required.  I don't particularly like all the layers but I dislike duplication more.

It sounds like the second suggestion would be the best place to start.  I think his main point on validation is don't interrupt the user when they are "in the flow" of completing the form and don't disable the "continue/submit" button.  These are the points that I've seen mentioned in lots of different places.  The other stuff might be considered more of a personal opinion and would seem to be less important.

There are still a few things I would like to do with the forms but is outside of this discussion e.g. remove placeholders, set up the forms semantically etc.

edwh:
Hold off until first submit and keep the components as they are.

The other stuff in the article is quite interesting and would be worth 
looking at further, but I can see us getting a bit bogged down in it.

JayyajGH:
> The other stuff in the article is quite interesting and would be worth
> looking at further, but I can see us getting a bit bogged down in it.

Indeed.  It would be great to make it all 100% perfect but it would never get finished that way.

nicksellen:
Thanks for the input, that's clear for me on how to proceed :)

nicksellen:
Ok, some progress :)

- I went with the behaviour to hold back on validation until they first press submit
- if there are errors we jump up to the first error
- I put things in a [`<b-form-group>`](https://bootstrap-vue.js.org/docs/components/form-group) so it also colours the label
- I made some attempt to handle resetting the form better - not sure what is best here, if you click outside the modal it closes, but I leave all the state as it is, if you click _Cancel_ it will reset the event and the data to initial values, doesn't seem perfect to me, but maybe enough for now
- have not tried editing existing events yet, as to try it I have to create an event and get it approved (annoying cycle when developing! would like a better dev environment to solve that)
- I managed to put validation around the group select and the start end collection without actually touching those components... (although visually the group select does not get coloured like the other input fields, but that's ok for now)

Here's a screenshot showing all fields with an error:

allerrors

Comments?

edwh:
Coming along nicely.  A couple of thoughts:

- I would be inclined to move the validation for start/end dates inside that component, with props to control behaviour if you need it.  It's that component's responsibility, I think, and otherwise you'll have to duplicate logic if we ever use that component elsewhere.
- Could you move any more code inside the mixin?  For example the focus on first error and the touch that you do on the submit looks like it would be fairly common code wherever we did this.
- @JayyajGH might have a view on whether, for accessibility reasons, we should include a b-alert when validation fails as well as highlighting the errors.

JayyajGH:
I think at the very least there needs to be something with a role="alert" to inform the user that something on the page has changed.  I think this is why the accessibility experts like the idea of displaying a list of all the errors on the page at the top of the screen.  It gives something to hook the alert role on to and takes the user back to the top of the page. Maybe even just a "There has been an error submitting your form.  Please resolve all the issues blah blah blah" message with a role of alert.

JayyajGH:
I saw somewhere recently that everyone should try and buy at least one Christmas present using a screenreader. I'm not sure I have that amount of free time!

nicksellen:
> I would be inclined to move the validation for start/end dates inside that component, with props to control behaviour if you need it. It's that component's responsibility, I think, and otherwise you'll have to duplicate logic if we ever use that component elsewhere.

In this case the validation of the dates is the responsibility of the community event (e.g. you need at least one), whereas for the volunteer things you can have none, but they still use the same component.

Also, I don't think vuelidate works if you try and have multiple nested components with independent validation (it needs to be on the main `$v` object, but the data can be arbitrarily deeply nested).

> Could you move any more code inside the mixin? For example the focus on first error and the touch that you do on the submit looks like it would be fairly common code wherever we did this.

It would be to create another mixin actually, the one on the CommunityEventModal is just the regular vuelidate one, and the one I created is just for the specific form components. I might make one that includes the vuelidate one, plus our extensions (like focus first error...). Seems a reasonable approach.

nicksellen:
This is the (slightly clean up) HTML that comes out of the bootstrap component when there is an error on an input field:

```
<div role="group" class="form-group is-invalid" aria-invalid="true">
    <label for="title" class="d-block" >What's the event's name?</label>
    <div>
        <div id="title" type="text" placeholder="Give the event a short title">
            <input id="title" type="text" placeholder="Give the event a short title" class="form-control is-invalid" aria-describedby="title-feedback" maxlength="80" aria-invalid="true">
            <div id="title-feedback" class="invalid-feedback"><span>Please add a title</span></div>
        </div>
    </div>
</div>
```

Is that enough `aria`? Or maybe adding the `role="alert"` to the error message?

JayyajGH:
Does vuelidate add an "aria-invalid".  Ha! You've just answered that question!  That's good.

Webaim have some good stuff to say about form validation.  Much of it we have already discussed though.
https://webaim.org/techniques/formvalidation/

JayyajGH:
I don't think you need to give every message a role=alert.  I think it tends to be just the main error.  In an ideal world that would be the list of all errors at the top of the page.

nicksellen:
That article says: _This error message should be visible, informative, and accessed directly. One method is to use a JavaScript alert box_ - I can't take it seriously after that :sob: 

Is it really better to put errors at the top in a list? I would hope `aria-invalid` somehow lets the user navigate between them? Errors at the top to me, _as well as_ inline with the fields seems kind of overkill... 

Btw the `aria-invalid` is from bootstrap, not vuelidate. Vuelidate does nothing regarding presentation, no components, html, error messages or anything, it only concerns itself with the logic of validation (to keep it flexible to use with whatever components you want).

JayyajGH:
Ha!  Everyone loves a javascript alert box!  Listing them all at the top does appear to be the solution listed on many different websites.

I would suggest that as long as the user is alerted to the error in some way (e.g. role=alert somewhere) and each field has an error state (e.g. aria-invalid) then that should be enough.
I've not got much experience with accessible validation so I am guessing a bit here.

nicksellen:
Ok, _maybe_ this is mergeable now.

A significant change I made after using it for _editing_ events, is to clone the incoming `event` prop into `eventEdit` data field. Otherwise we end up mutating the contents of the `event` prop, and the listing of the event gets updated as we type.... Actually, I like this editing model in general, as it lets us keep the original event around for comparing to (e.g. the `olddates`...) logic.

Also, I moved the logic to focus the first error field into a mixin.

There are a couple more TODOs in there, but maybe useful to have the validation in the main code?

I might have a few days before I can do more on this again (Monday+ probably). Happy if you want to leave it to after then to merge too :)

edwh:
Thanks.

I think it's mergeable as is so I'll do that this evening to get it in.

Minor comments for when you get back to it:
Will placeholder work?  You don't list it as a prop for your new 
input/textarea components.  Ditto minlength, spellcheck, probably 
others.  Unless you've done something fancy to inherit all of them which 
I can't spot because the guy next to me has his headphones on too 
loud."It's id" should be "Its id" (sorry).  "detirmined" => 
"determined"Do we really need a separate red colour?  We have too many 
colours so I would be inclined to make color-red and color-danger the 
same.There is obviously more validation to do on events.  Start/end 
dates (end after start, no event longer than 4 days), sensible min 
length on required fields, email syntax.  But you know this :-).Would be 
nice to have some more comments on the vuelidate stuff for those like me 
who aren't familiar with it, e.g. touch, dirty, anything on $v.The 
nextTick stuff looks deeply hokey and fragile - exactly the kind of 
thing I'd write and then feel it meant I wasn't using Vue properly.  If 
we really have to do this then I'd think a 1 second timer might be more 
robust.Some nice code-level tidyups in here, thanks.  I've not seen 
const { id } =this.event before; presumably that resolves to undefined 
if the property isn't present.  Personally I prefer to explicitly set 
things to null when I know they are absent and tend to reserve undefined 
for bugs, but that's personal preference.

nicksellen:
> Will placeholder work?  You don't list it as a prop for your new input/textarea components.  Ditto minlength, spellcheck, probably others.  Unless you've done something fancy to inherit all of them which I can't spot because the guy next to me has his headphones on too
loud.

Yup, there is something "fancy", the pattern is called "higher order component" (basically, one that wraps another one, in a more-or-less transparent way), and in vue you can do `v-bind="$attrs"` to pass down everything that wasn't declared as a prop (so the attrs on the component are a mix of props and attrs really). See https://vuejs.org/v2/api/#vm-attrs for a bit more detail.

The downside is worse auto-completion for IDEs :/

> Do we really need a separate red colour?  We have too many colours so I would be inclined to make color-red and color-danger the same.

I didn't add a seperate colour, just stole the definition from the bootstrap colours, but currently they are not imported/combined with `color-vars`. So, I had to re-declare it. bootstrap danger is the same as bootstrap red anyway. Would be nice to have access to bootstrap vars inside the custom freegle vars, but seems out of scope for this PR, maybe @JayyajGH has a thought on that?

> Start/end dates (end after start, no event longer than 4 days), sensible min length on required fields, email syntax.  But you know this :-).

Yup, was gunna wait until the work done to set end depending on start was done, as it is kind of related/interleaved. I think it shouldn't be needed to validate on that as the component will already ensure that is valid. But yes, for "event length", I'll add a TODO.

> "It's id" should be "Its id" (sorry).  "detirmined" => "determined"

Will fix :)

> Would be nice to have some more comments on the vuelidate stuff for those like me who aren't familiar with it, e.g. touch, dirty, anything on $v.

I would just refer you to the vuelidate docs ---> https://vuelidate.js.org/#sub-v-values

> The nextTick stuff looks deeply hokey and fragile - exactly the kind of thing I'd write and then feel it meant I wasn't using Vue properly.

Agree, it's a hack really. Ideally you don't need to think about which elements to trigger focus on, I think there should be a better way, but couldn't come up with something yet. Could set refs for each field and trigger focus like that - then don't need to rely on the bootstrap classes being set for referencing them. I think that should work. Another TODO!

I think it's reasonably deterministic though, I think there are a number of things that get triggered to re-render, i.e. 3 ticks never works, 4 ticks always works (within the confines of my playing around).

> I've not seen const { id } =this.event before; presumably that resolves to undefined if the property isn't present.  Personally I prefer to explicitly set things to null when I know they are absent and tend to reserve undefined for bugs, but that's personal preference.

The property is (aka should be) always present as we set `id: null` in `initialEvent`, but yes it would resolve to `undefined` if it wasn't present. But by creating a null valued property by default we allow the reactivity to watch it when/if it gets set to something.

JayyajGH:
> Would be nice to have access to bootstrap vars inside the custom freegle vars, but seems out of scope for this PR, maybe @JayyajGH has a thought on that?

Not entirely sure what you mean here?  Do you just mean you want to access the bootstrap variables?  I've done something similar somewhere here using:
`@import 'node_modules/bootstrap/scss/variables';`

To make something clear, I'm not a big fan of what I've done with the colour variables.  Ideally we would use css custom properties but to make that work in IE would need a polyfill/ponyfill and I've not used any before so that needs looking into.  At least we now have a common place to define the brand colours and it exposes how many there are which means we can start consolidating them.  Once we moved to custom properties it will then be easy to switch the current definitions over.  Consolidating the colours will mean checking where each one is used and on which colour background to check the contrasts are AA compliant before we remove some.

nicksellen:
OK, I switched to a ref-based approach for focusing the first error. More proper, but also needs more fields to be set (namely, the ref field), could consider abstracting that away too in the future...

nicksellen:
@JayyajGH more concretely, when using the bootstrap `:state` prop on form components it puts a nice red/green border and label depending on validation status.

In `CommunityEventModal` though the label colour is set to `$color-green--darker` which override the nice red from bootstrap. I wanted to keep the green for normal case, but make it the same as the border for error cases, but `_color-vars.scss` does not include the bootstrap colours, so I didn't have those available. I didn't really want to pull in all the bootstrap vars, as it seemed having our own color vars is useful.

Actually I didn't even want to touch any of that... 

I thought to switch to importing bootstrap vars, but it gives me this big fat error:



```
SassError: argument `$color` of `darken($color, $amount)` must be a color
        on line 174 of node_modules/bootstrap/scss/_variables.scss, in function `darken`
        from line 174 of node_modules/bootstrap/scss/_variables.scss
        from line 315 of /code/freegle/iznik-nuxt/components/CommunityEventModal.vue
>> $link-hover-color:                        darken($link-color, 15%) !default;
   ------------------------------------------^
```

... even though the import is used elsewhere (fyi, you can use `@import '~bootstrap/scss/variables';` so don't have to have the ugly `node_modules` bit there).

nicksellen:
... aha I had to import bootstrap functions too :)

JayyajGH:
Yeah I made that mistake too!  Generally I like the idea of colours being defined in one place.  That way you get a consistent look and feel across the whole site.  I don't mind where those colours are defined but I think there needs to be a place to go that lists them all.  I think going forward it would be good to have all freegle colours listed in the same place and not define extra colours outside of that.  I think maybe stuff used from bootstrap needs to be dealt with separately. It might just be the case of adding a comment in the colour definition file maybe.  I'll have a think...

Ah yes "~". Is it nuxt that defines that alias?

nicksellen:
> Ah yes "~". Is it nuxt that defines that alias?

According to https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules it tells webpack not to treat it as a relative import, which makes it look in node_modules... most importantly it's different to `~/someLibrary` in js files (which means relative to the project root).

PR #77, opened 2019-12-04
Set the initial height of the sidebar items

JayyajGH:
Resolve the TODO to make the sidebar items equal height.  When there is only one present then it should still only take up 50% of the space.

To do this I'm using flex-basis to set the initial size of each sidebar item to be 50% of the sidebar height.  I've also set flex-grow to be 0 so if there is only one item then it doesn't fill the whole 100% of space.  flex-shrink is left at the default.

### DIFF
--- a/components/SidebarRight.vue
+++ b/components/SidebarRight.vue
@@ -1,8 +1,7 @@
 <template>
   <div class="d-flex flex-column sidebar__wrapper">
-    <VolunteerOpportunitySidebar v-if="showVolunteerOpportunities" class="justify-content-start flex-grow-1" style="overflow-y: auto" />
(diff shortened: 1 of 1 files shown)


edwh:
Thanks.  

I worry a bit about the 100px that's in there from earlier.  One TODO I have somewhere is whether to do that thing where you shrink the navbar as you scroll down, for example.  And it assumes it's the same height on all device sizes.  Are we able to do anything cleverer?

JayyajGH:
Yeah stripping off the height of the nav bar from the sidebar is a bit messy. 100vh is also a bit awkward too as it doesn't play well with mobiles.  Basically it doesn't take into account the browser bar popping up so the height ends up being bigger than the screen.  Maybe just add some TODOs with what you want and I'll take a look.  Is there the option of removing the sidebar for example and displaying that information elsewhere?

PR #79, opened 2019-12-06
Add visibility watcher

nicksellen:
A first pass at a visibility watcher, re: https://app.slack.com/team/U6MA8D84T

A few random thoughts:
- maybe should include "time since last visible", as you might not want to _always_ refresh things, if it is heavy to do so
- perhaps we don't need the event timer based detection (that detects idling on the page, like maybe you left the window open for some time, then came back) - I'm not sure why we added this in karrot, maybe to support browsers that don't support visibility API (though looks not too bad, https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API#Browser_compatibility), or maybe some edge cases :/
- I made it trigger `notifications/count` and `notifications/list`, but not sure if that is correct

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -79,6 +79,7 @@ module.exports = {
     { src: '~/plugins/dayjs'},
 
     // Some plugins are client-side features
+    { src: '~plugins/visibility.js', ssr: false },
     { src: '~plugins/error-toasts.js', ssr: false },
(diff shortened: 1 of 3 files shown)


edwh:
I've merged this; I'll tweak what it calls. 

The "time since" part might be nice, but I don't think people shifting away and back again very rapidly is so likely to happen that we need to optimise for that.

PR #82, opened 2019-12-10
Update the login form fields to have an associated label.

JayyajGH:
So I've backtracked a little on the stuff I was doing on the login form.  There are potentially a lot of questions etc. in that area so I thought I would strip it down to basics and do a step at a time.

This change just updates the current input field with placeholder into an accessible input field with associated label.  Are you happy with this style?  The benefits you get here are:
- The field works correctly with a screenreader
- You can click the label to focus on the field which is great for people with motor issues
- Placeholder text isn't automatically translated by browsers; the label is.

Do we need to keep the pre-pended "Your" on the fields? i.e. "Your email address", "Your password", "Your name" etc.  My personal opinion is that I think it makes it harder to scan for the words you are used to like "Password" and "Email" but you know your clients better than I do!

I've not actually tested the functionality of this yet - I've just updated the visuals although I'm pretty sure it will work.

### DIFF
--- a/components/LoginModal.vue
+++ b/components/LoginModal.vue
@@ -54,59 +54,79 @@
           <div>
             <b-row v-if="showSignUp">
               <b-col>
-                <b-form-input
-                  id="firstname"
-                  ref="firstname"
-                  v-model="firstname"
-                  name="firstname"
-                  placeholder="Your first name"
-                  alt="First name"
-                  class="mb-3"
-                  autocomplete="given-name"
-                />
+                <b-form-group
+                  id="firstnameGroup"
+                  label="Your first name"
+                  label-for="firstname"
+                  label-class="mb-0"
+                >
+                  <b-form-input
+                    id="firstname"
+                    ref="firstname"
+                    v-model="firstname"
+                    name="firstname"
+                    class="mb-3"
+                    autocomplete="given-name"
+                  />
+                </b-form-group>
               </b-col>
             </b-row>
             <b-row v-if="showSignUp">
               <b-col>
-                <b-form-input
-                  id="lastname"
-                  ref="lastname"
-                  v-model="lastname"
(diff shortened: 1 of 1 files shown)


edwh:
Happy with this style for the good reasons you give.

The reason I have "Your" is because I find "Email address" quite cold 
and unwelcoming, like something barked at you by an East German border 
guard.  So I'd prefer to keep that unless you can think of some other 
way of making the modal feel friendly.

JayyajGH:
OK cool.  I'll give it a test and submit the PR.  So many hugely popular sites still do this so badly.  Probably not for much longer if Dominos lose the "Domino's v Robles" accessibility case though!

Yeah I can see what you mean.  My only other comment on the "Your" would be the sites that people use the most like Facebook, Amazon, Instagram, BBC etc. all go for the border guard style.  People are used to it.  I'll go ahead with it as it is for now unless you change your mind!  Maybe we could put a kitten on it? :-)

edwh:
Kittens.  Of course.  Mock that up?

JayyajGH:
This is one of my favourite image placeholders:

[](https://placekitten.com/)

edwh:
I deduced the existance of and then located this: 
https://vole.wtf/plaiceholder-images/

JayyajGH:
I'm not getting any feedback if I enter invalid login details though.  Is this part of the work that Nick is doing or is this a separate issue?

edwh:
I pushed a fix for that yesterday, I think - have you merged recently?

JayyajGH:
Ah no sorry I haven't merged since yesterday.  Should have merged before submitting the PR....

JayyajGH:
This is now working ok. :-)

edwh:
Good.  Normally I don't have as good an answer to bugs as that...

PR #85, opened 2019-12-11
Add validation for volunteer ops

nicksellen:
Ok, this makes it very similar to the community event validation.

It's quite hard to check if it's all ok, I created a volunteer op, but needs to be approved....

I moved the shared form styles into the global scss file, that involved import bootstrap functions and variables, what do you think @JayyajGH?

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -328,19 +328,3 @@ legend {
 .mx-datepicker-popup {
   z-index: 2000 !important;;
 }
-
-label,
-.col-form-label {
-  font-weight: bold;
-  color: $color-green--darker;
-  margin-top: 10px;
-}
-
-.is-invalid label,
-.is-invalid .col-form-label {
-  color: $form-feedback-invalid-color;
-}
-
-.form-group.is-invalid .invalid-feedback {
-  display: block;
-}
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -1,5 +1,3 @@
-@import '~bootstrap/scss/functions';
-@import '~bootstrap/scss/variables';
 @import 'color-vars';
 
 // TODO Would be nice to use bootstrap mixins for breakpoints.  Also in NewsReply.
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -1,3 +1,5 @@
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
 @import 'color-vars';
 
 // TODO Would be nice to use bootstrap mixins for breakpoints.  Also in NewsReply.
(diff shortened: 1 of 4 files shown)


edwh:
I've approved the volunteer op.

JayyajGH:
Sorry for the delayed answer @nicksellen . Totally missed your tag.

So this is just my personal opinion so feel free to ignore!

I think the global scss file should just be for truly global stuff.  Setting the fonts, header style, generic padding and spacing, buttons etc.

It's a simple site but this is a global file I created for another site.
https://github.com/JayyajGH/brendans-brewtique/blob/master/src/assets/css/global.scss

If there are styles shared by forms then they should probably live either in a shared form component/mixin/thing or in a a a component directly for the item e.g. text entry, date entry etc.  They are only used on a subset of pages and so they are not global.  Not sure how feasible that is with the work you are doing though.

I don't really like the bootstrap import stuff but I figure it's only a build time thing so I'm ignoring it.  A longer term solution might be just to use CSS custom properties which would make things much simpler - but no native IE11 support.

Excuse the colour definition naming at the moment btw.  It was just the first step of consolidating the colours into a manageable set.  I think the colours need reducing by at least 50% which will then make naming them much easier.

nicksellen:
Thanks!

So, I see two options:
1. define `_forms.scss` that is then imported into each component that uses those styles
2. define a Form component that contains the styles and wraps any forms

Any preference (or other option)?

JayyajGH:
Generally I think anything that keeps the styles defined only in the places that they will be used would be a good thing.

My preference would be for some kind of component.  If you have forms in multiple places all including the same HTML and styling classes then abstracting that out would keep the HTML DRY and provide somewhere that the classes could live too.  The presence of a _forms.scss file would be a bit of a smell for me that something needs abstracting.  Although I have done something similar in the past where another component just wasn't right.

nicksellen:
This is just about a form wrapper really, rather than all the HTML, the HTML comes from the bootstrap form components, so would be something like:

```html
<some-form-wrapper>
  <b-input-or-whatever></b-input-or-whatever>
</some-form-wrapper>
```

The form wrapper would be outputting a form tag, and include the styles, but nothing else as the insides of the form are other components rendered as children.

I think there is potential for further abstraction at some point (setting the refs automatically, etc...), but I'm going step by step.

(_slight aside_: to my logic, form styles are as global as any of those other things, not _every page_ has a primary button on either).

JayyajGH:
> (slight aside: to my logic, form styles are as global as any of those other things, not every page has a primary button on either).

I think that shows that this topic is definitely a personal preference thing!  I can totally understand your logic.  I guess it depends on the context of the website.
100% forms: Global.
5% forms: Component.
6-99%: Personal preference.

I think in this case I would probably create a component for it. Especially if if seems like there is potential for more abstraction.  At the very least you will benefit from the scoped styles.

As a newcomer to this project the thing that confused me the most was the huge number of styles in global.scss and I didn't have a clue where any of them were used.  Especially as some were styling third party components so the classes weren't in the code.  If they are in a component then you know exactly where they are used. You might also be able to reduce the specificity of some selectors too with a smaller context.

edwh:
Just to be explicit: I have nothing to add to this discussion :-).

nicksellen:
I created a component for it :)

`<validating-form>` ... it's name is a bit of a lie, as it doesn't really do anything itself, but it's named to show it's part of the other `<validating-*>` components. And the style is now scoped inside that.

PR #86, opened 2019-12-11
Enhancements/update social logos

JayyajGH:
Update the social media login buttons to be consistent.

They are now buttons so will be accessible too.  I've kept most of the classes in CSS rather than using bootstrap ones.  Because quite a few of the attributes couldn't be set from bootstrap like the widths and colors, it seemed quite messy to have a split between where they were defined.  If you would rather I defined as many as I can in the HTML using bootstrap then I'll update them.

The buttons don't seem to work properly using localhost. Facebook and Google have oAuth issues and Yahoo doesn't do anything. It did that to me before this change and after this change so you might want to give this a test to check it does what you would expect.


edwh:
Thanks.  There are issues on localhost, but I'll test it on the deployed preview site.

Not fussed about the CSS.

edwh:
Looking at it once merged, and now there are the labels, I wonder if it would look neater if you could align the top of the first social button with the top of the first input on the right hand side, and the bottom of the bottom social button with the bottom of the Freegle button.  

The following picture isn't quite aligned but is closer, and shows what I mean:

image

JayyajGH:
Yup agreed. There is also too much space in the middle of the screen too. I was going to sort out the alignment with the stuff to resolve the other TODO to have a better split between the two sides.  The only issue with alignment here is when the signup options are also present.  I'll try and think of a solution...

PR #88, opened 2019-12-12
Create a more obvious divide between the two sections.

JayyajGH:
This PR is to try and create a more obvious divide between the two main login options.  This works across all breakpoints.

I've not yet thought of a way to align the social buttons with the RHS any better.  Still thinking...

I've tweaked the bootstrap modal header element as it was creating the heading as an h5 element.  One of the rules of accessibility is that header levels should be in order and not missing any levels.  The heading is now an h2 with the subheadings as h3.  Bootstrap seems to directly link the semantic heading elements with the visual style which is probably why it's like that.  The main page is actually missing an h1 but that is another issue.

I've also removed a lot of the row and column elements.  In this case they weren't providing anything as the standard elements do what is required by default.  Divs are by default flexible width elements that take up 100% width of their container.

Thoughts?


JayyajGH:
I've also tried to replicate the facebook language of "Continue..." as it works well for the headings, especially as they can refer to both signing in and signing up.

edwh:
The divider itself looks fine; perhaps "OR" is a little faint.  The positioning looks a bit off - it's closer to the right hand side than the left.

I think "Continue with your Freegle account" is ok on Sign In, and is consistent with the social buttons.  But on Sign Up, you can't continue with something you don't yet have (whereas you do already have the social accounts) so that looks weird.  Perhaps "Create an account on Freegle"?

The standard bootstrap H1 is huge, and there are definitely some places where I miss levels because the next-level heading felt too big in context.  I'm not sure what to do about that.  What is the accessibility impact of missing levels?

JayyajGH:
They are actually exactly the same width!  I think it's just an optical illusion.  I could move it further to the left though if you want.  Each side is currently 45% and justified with "space between".

Yup good point re. "create an account".  I'll switch that over.  And update the colour for the divider.

JayyajGH:
Fundamentally the heading levels are semantic, not visual.  You should be able to remove all of the page styling and the page still make sense.  I tend to think of it as chapters in a book.  You have the main title of the book (h1) and then the chapter names (h2) and then headings within those chapters (h3).  The size, colour, weight etc. has got nothing to do with the HTML element.

I believe one of the main ways of navigating with a screenreader is to skip between headings.  If there are some missing or they don't make sense then it will confuse the client.  So using this page as an example.  You know the main heading of the modal.  If you couldn't see the screen then what would you want to see next?  Probably see what the main options are which are "Continue with social media" or "Continue with freegle account". 

https://www.nomensa.com/blog/2017/how-structure-headings-web-accessibility

JayyajGH:
I've now updated the freegle account text and made the OR slightly darker.  Shout if you think it should be darker still.  I like it being fairly subtle but obviously it still needs to be seen!

edwh:
The sides might be the same, but the contents of each side are not.  So 
the result is that there is a bigger gap on the left hand side because 
the buttons are narrower than the text on the right hand side.  That 
means there's more whitespace, which throws it off balance.  See what I 
mean?

JayyajGH:
I've evened out the two sides a little better now.  It's quite tricky as I'm trying to keep the headings on one line and also the bootstrap modal is fixed width so you can't just leave flexbox to do it's thing.

edwh:
Sorry, I hadn't noticed this was ready for review.  Could you have a look at the conflicts before I do?

JayyajGH:
Ah yes, I'd forgotten about it too.  I'll sort them out and give you a shout!

JayyajGH:
Right that should now be done.  That was a horrible merge!

I've also added a couple of TODOs as well.

edwh:
I'm sure that was a nasty merge, we'd both been in there.  Released now, and I fixed the eye TODO.

JayyajGH:
Cool.  One thing I was trying to think about was the show/hide password icon usually gets in the way of my password manager icon but for some reason this one seems to work. Nice!

edwh:
Probably because it's a separate element next to the input, rather than 
displayed on top of it, so your password manager icon will focus on the 
input.  The same reason that the shadow stuff isn't quite right.

JayyajGH:
Ah right. Yes that sounds right.  The focus indicator will be tricky to get right here.  The eye icon needs to be able to be focused separately by keyboard and have an obvious indicator.  But I agree that it does look a bit weird at the moment.

PR #91, opened 2019-12-18
Refactor the top section to use flexbox rather than the bootstrap grid

JayyajGH:
First part of the chitchat page tidy to resolve the width issue.  I thought I would do this in small increments.

Refactor the top section of the page to use bootstrap flexbox rather than the bootstrap grid.  This also resolves the issue with a button contained within an anchor which was creating double button tabbing.

### DIFF
--- a/components/CommunityEventSidebar.vue
+++ b/components/CommunityEventSidebar.vue
@@ -18,7 +18,7 @@
         </div>
       </b-card-body>
     </b-card>
-    <CommunityEventModal ref="eventmodal" :event="{}" :start-edit="true" />
+    <CommunityEventModal ref="eventmodal" :start-edit="true" />
   </div>
 </template>
 <script>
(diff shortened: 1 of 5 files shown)


JayyajGH:
The tooltips that pop up appear to have issues scrolling up sometimes.  Not sure if this is to do with the fixed toolbar.  It works the same way it did before this change but you might want to take a look and add another todo.

edwh:
Which tooltips?

JayyajGH:
The ones that appear if you enter any "find" or "give" type words in the chitchat box.

edwh:
Ah, ok.  Can you get a screenshot of what you mean?

JayyajGH:
capture4

JayyajGH:
I think it only happens if the button is offscreen so it's probably not the end of the world.

edwh:
Ta - pushed a fix for that.

JayyajGH:
Awesome!

PR #92, opened 2019-12-19
General accessibility improvements

aaronferris:
Quick accessibility improvements, it's generally good practice to always include hover and focus in the same ruleset for keyboard users.


JayyajGH:
I like accessibility improvements. :-)

PR #93, opened 2019-12-20
Add a bootstrap container to house the rows and columns

JayyajGH:
Bootstrap rows and columns should live in a bootstrap container.

https://bootstrap-vue.js.org/docs/components/layout/

### DIFF
--- a/pages/chitchat/_id.vue
+++ b/pages/chitchat/_id.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="me">
+  <b-container v-if="me" fluid>
     <b-row class="m-0">
       <b-col cols="0" lg="3" class="d-none d-lg-block">
         <SidebarLeft :show-community-events="true" :show-bot-left="true" />
@@ -128,7 +128,7 @@
         <sidebar-right show-volunteer-opportunities show-job-opportunities />
       </b-col>
     </b-row>
-  </div>
+  </b-container>
 </template>
 
 <script>


edwh:
Does this fix the margin issue, as per https://stackoverflow.com/questions/31885727/is-it-acceptable-to-omit-container-elements-when-using-bootstraps-grid ?

If so, then there will be quite a lot of other examples of this mistake.

JayyajGH:
Nope not yet.  This was a standalone change so I thought I would get it in.

My general plan for this file is:
- Add container
- Move area options out of chitchat text box.  This will make it more related to where it is used and sort out the alignment on smaller screens.
- Sort out the photo box sizing
- Tidy up the rows and columns left behind - This will resolve the margin issue. I think the rows and columns just add unnecessary complexity here. Just making them bootstrap flexboxes will achieve the same effect with much less code and with none of the negative margin mess. Using negative margins to sort out the left and right hand spacing is one of the main issues that flexbox was designed to solve.

I also think that maybe creating a component with a slot to house this common layout which is definitely used in at least three separate components.  That will resolve the issue in a bunch of other places too.

edwh:
All sounds good.  I'll merge this one.

JayyajGH:
It is quite convoluted way to get to the actual fix for the issue but I think it will tidy things up a lot on the way.  Honestly!

edwh:
I believe you.  I've done a little fix for this issue, but don't let 
that put you off this, as it's a worthwhile change that will help stop 
me reaching for b-row all the time.

JayyajGH:
You fix doesn't appear to work for me btw.  It fixes the top section but the problem then occurs in the newsfeed section itself now.

edwh:
Ah, ok.  Might look further then, might not.

JayyajGH:
It's the NewsUserIntro that is causing problems for me now.

PR #95, opened 2019-12-20
Enhancements/newsaboutme refactor

JayyajGH:
So this change is the one that fixes that right hand gap on mobile. :-)

It's basically removing a lot of the rows and columns from the various newsthread components. I've currently only changed the places that were causing problems.  There also might be more depending on what is in your newsthread.  I'll have a play this afternoon to see if I can find any more.  But this one is good to go for now.


edwh:
Well, that was a larger fix than you might have expected it to be...

JayyajGH:
Yeah indeed!  I'll continue to have a poke around in this area to see if anything else is causing issues.  Although that is likely to be after Christmas now!

PR #97, opened 2019-12-20
Menu for small screen improvements

aaronferris:


aaronferris:
As discussed and showcased on Slack

PR #98, opened 2019-12-20
Resolve icon hover/active state inconsistencies

aaronferris:


layouts/default.vue
```diff
@@ -301,11 +301,27 @@ nav .navbar-nav li a.nuxt-link-active[data-v-314f53c6] {
...
+  }
 }
 
-.navbar-dark .navbar-nav .nav-link:hover,
-.navbar-dark .navbar-nav .nav-link:focus {
-  color: $color-white-opacity-75 !important;
+#menu-option-chat-sm {
+  &:hover,
+  &:focus {
+    color: $color-white-opacity-75 !important;
```
aaronferris: Specificity wars :( :( :( 

Im assuming we're using !important alot in our codebase because of bootstrap styles? This causes verbose styles, changing at this stage is going to cause a risk of regression however.

assets/css/global.scss
```diff
@@ -182,35 +182,35 @@ body {
 
 // Brought in from style.css
 button {
-    white-space: nowrap;
+  white-space: nowrap;
```
aaronferris: Indentation changes for consistency.....

layouts/default.vue
```diff
@@ -301,11 +301,27 @@ nav .navbar-nav li a.nuxt-link-active[data-v-314f53c6] {
...
 .navbar-dark .navbar-nav .nav-link {
   color: $color-white !important;
+
+  &:hover,
+  &:focus {
+    color: $color-white-opacity-75 !important;
+  }
 }
 
-.navbar-dark .navbar-nav .nav-link:hover,
```
aaronferris: Not removed, nested above for cleanliness

edwh:
I think many of the uses of important are wrong. But as you say, bit risky to change just now.

PR #100, opened 2019-12-22
Gauge prefix

aaronferris:
This does three things

1. Gives us an option to patch node_modules as part of a build - this could well be handy for further usage, as I think moving the 'scale' as a patch in the plugin wouldnt really make sense, given the use case
2. Moves the gauge suffix to be a prefix
3. Adds a £ to the scale

Other options would be:

1. Use a different plugin
2. Rebuild this template to use the different plugin

This does feel a little 'hacky' to me, CSS isnt an option due to the way the gauge is an SVG, the usage of the patch package is a nice quick clean option imo.


patches/vuejs-thermometer+0.0.4.patch
```diff
@@ -0,0 +1,34 @@
+diff --git a/node_modules/vuejs-thermometer/.DS_Store b/node_modules/vuejs-thermometer/.DS_Store
```
aaronferris: Need to remove this

nicksellen:
I'm not in favour of using patch-package in general, as it can break on upgrades easily. I don't have any other ideas (or time really) though. Maybe it'd be possible to add an option to the upstream library?

edwh:
I didn't know about patch-package - thanks.

Probably for small ones like this we should do a pull request on the original, so that even if it doesn't get merged it's there for other people to grab if they need the same function.

I have a beam in my own eye, as I've taken a couple of packages into local ones in this project and modified them.  I should review those and submit PRs where appropriate.

PR #102, opened 2019-12-22
Login modal on < 992 device width central social media buttons

aaronferris:


JayyajGH:
If we're generally working mobile first then we should probably set centre as the default and have left aligned as the media-breakpoint-up value. If that makes sense...

JayyajGH:
👍

PR #105, opened 2019-12-27
Tooltip opacity

aaronferris:


assets/css/global.scss
```diff
@@ -304,3 +304,8 @@ legend {
 .mx-datepicker-popup {
   z-index: 2000 !important;;
 }
+
+/* Overrides default Bootstrap tooltip opacity (0.9) */
+.tooltip.b-tooltip {
```
aaronferris: Feels a bit odd why Boostrap would set an opacity on something like this to 0.9 in the first place....

PR #106, opened 2019-12-27
Make the /spread page white background by default

aaronferris:
Not sure what else to suggest for this TODO item, imo anything else is going to need a large site wide redesign in oder to incorporate the wallpaper and some sort of white wrapper around pages that are text heavy. I dont personally see anything wrong with the approach we're already using on other pages (like /help) - by default pages have the wallpaper, override where required.

As above, anything else feels like a site wide redesign with a white background container (set width) that sits on top of the wallpaper.

I do think this page needs some attention though, perhaps some sort of accordion?

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -104,9 +104,6 @@ div[variant="success"] .card-header {
 }
 
 body {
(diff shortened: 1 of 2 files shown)


edwh:
Yes, I think bg-white where it looks better is the way to go.

PR #112, opened 2019-12-28
Remove tablet restrictions on columns for the find process

aaronferris:


edwh:
I'll merge this and the others, but they have stuff like this:

```
      <b-col cols="0" md="3" class="d-none d-md-block" />
      <b-col cols="12" lg="6" class="p-0">
```

I think we need to change the first one as well as the second.

aaronferris:
> I'll merge this and the others, but they have stuff like this:
> 
> ```
>       <b-col cols="0" md="3" class="d-none d-md-block" />
>       <b-col cols="12" lg="6" class="p-0">
> ```
> 
> I think we need to change the first one as well as the second.

Shouldnt matter as its basically an empty div, we could tidy it up though

edwh:
I'd expect it would mean that on md size we'd have 15 columns and 
therefore wrap.  I took a pass through and removed the ones I could 
find.

PR #120, opened 2019-12-30
Enhancements/add profile image component

JayyajGH:
This is the first stage of creating the profile component to sort out some of the broken image issues.  There are quite a few parts to this - updating quite a few components and moving some styles around - so I thought I would check it in a bit at a time so the change doesn't go stale before I finish it and become a merge nightmare. 
If you're happy with this change then go ahead and merge it and I'll work through using this everywhere as time permits.


edwh:
Do we need the name property on the component?

You're setting a class of profilesm/profilemd on this component.  I 
think it would be better to indicate the size via a size prop, as 
bootstrap does and we do in some other components, and then apply the 
relevant class within the component.

JayyajGH:
Many moons ago I read somewhere that every component should have a name property set.  That might have changed now though...  I'll look into it...

Yup that would be a good plan.  When I did this initially I just tried to keep the changes to a minimum but I'll update that.  There are probably a number of other things that could be refactored out too which will be clearer to see once all places are using the component.

JayyajGH:
I think this is where I read about always adding a name to a component:
https://vuejs.org/v2/api/#name

JayyajGH:
Looks like there is a current PR to add a check for a name property to eslint too...

https://github.com/vuejs/eslint-plugin-vue/pull/945

edwh:
Good spot on the name.  I'd better do that, I guess.

edwh:
Is this still WIP or waiting for merge?  I'm not sure if you did the size prop.

JayyajGH:
I've not done the size prop yet.  I have work work to get done today so I'll take a look on Friday.

JayyajGH:
Ok @edwh - This should now be ready to merge.  I'll then sort all the others out on Friday.

PR #154, opened 2020-01-02
130 sidebar visual improvements

aaronferris:


components/SidebarLeft.vue
```diff
@@ -28,10 +28,6 @@ export default {
 </script>
 
 <style scoped>
-.sidebar__wrapper {
```
aaronferris: Im not sure what our preference is for this sort of thing component Vs Global css but for me, if it's a reusable piece of code it should sit in global to be reusable and editable in a single place.

nicksellen: An approach I used elsewhere, I forgot where it was, is to create a component that they can both use. Maybe a SidebarWrapper?

JayyajGH: My general thought here is that we could create a generic sidebar component that could house a slot for the content.  But the sidebar is used in multiple places so we'd then be duplicating the slot content.  But each of the places that uses it is also just duplicated as well so we should probably create a layout component for chitchat, communities and myposts which could then set up the sidebars.  But that all obviously just expands the work out even more so probably best to ignore for this change and deal with separately.

Not sure if that made any sense!

JayyajGH: So most of my sites tend to have a layout file looking something like this:

<template>
  <div>
    <site-header></site-header>
    <nuxt />
    <site-footer></site-footer>
  </div>
</template>

That way we could then easily create different layout styles for the ones with and without sidebars. The current setup doesn't really allow us to do that.  That would then allow us to create a common sidebar component that would resolve this issue.  One sidebar in one location containing the appropriate styles.

I might create a separate discussion for this...

JayyajGH: Issue #155 created

edwh:
So this is improved a bit, but still has the sharp cut-off at the bottom, which we've been discussing whether/how to do anything about.

PR #157, opened 2020-01-06
Enhancements/use profile image

JayyajGH:
The next round of using the profile image component

### DIFF
--- a/components/MyMessageReply.vue
+++ b/components/MyMessageReply.vue
@@ -44,7 +44,7 @@
             </v-icon>
             Unpromise
           </b-btn>
-          <b-btn v-else variant="success" class="align-middle mt-1 mb-1" @click="promise">
+          <b-btn v-else-if="message.type === 'Offer'" variant="success" class="align-middle mt-1 mb-1" @click="promise">
             <v-icon name="handshake" /> Promise
           </b-btn>
           <b-btn variant="primary" class="align-middle mt-1 mb-1 mr-1" @click="chat">
--- a/components/MyMessageReply.vue
+++ b/components/MyMessageReply.vue
@@ -2,23 +2,13 @@
   <div>
     <b-row>
       <b-col cols="6" class="text-truncate">
-        <span class="">
-          <b-img-lazy
-            rounded="circle"
-            thumbnail
-            class="profilesm p-0 ml-1 mb-1 mt-2"
-            alt="Profile picture"
-            :src="reply.user.profile.turl"
-            @error.native="brokenImage"
-          />
+        <span>
+          <profile-image :image="reply.user.profile.turl" class="mr-1 mb-1 mt-2 inline" is-thumbnail size="sm" />
           <span v-if="unseen > 0" class="align-middle">
             <b>{{ reply.user.displayname }}</b>
           </span>
           <span v-else class="align-middle">
             {{ reply.user.displayname }}
-          </span>
-          <span class="align-middle text-muted">
-            last wrote to you:
           </span>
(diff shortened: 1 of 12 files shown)


components/MyMessageReply.vue
```diff
@@ -161,3 +134,22 @@ export default {
...
 }
 </script>
+
+<style scoped lang="scss">
+@import 'color-vars';
+
+.snippet {
+  border: 1px solid $color-gray--light;
+  border-radius: 10px;
+  padding-top: 2px;
```
aaronferris: Could short hand this

components/ProfileImage.vue
```diff
@@ -77,4 +77,14 @@ export default {
     height: 50px;
   }
 }
+
+.profile--xl {
+  width: 100px;
+  height: 100px;
+
+  @include media-breakpoint-up(md) {
```
aaronferris: Feels slightly odd to have a media query setting the same value as the default

pages/settings/index.vue
```diff
@@ -824,3 +796,19 @@ export default {
   }
 }
 </script>
+
+<style scoped lang="scss">
+@import 'color-vars';
+
+.groupprofile {
+  height: 100px !important;
```
aaronferris: Can we avoid the !important?

pages/settings/index.vue
```diff
@@ -824,3 +796,19 @@ export default {
...
+}
+
+.nocardbot .card-body {
+  padding-bottom: 0px;
```
aaronferris: Dont need to specify 'px' just '0' will do.

JayyajGH:
@aaronferris - At the moment I've just replaced the profile image code with my new component.  I haven't refactored any of the old existing code in the components I've changed.  Although I agree with your comments I'm sort of loathe to make lots of random changes and increase the scope as part of this PR.  I don't even know if half of those old classes are still used.

I'll look at the media query one for the XL breakpoint as I think that is incorrect.

components/ProfileImage.vue
JayyajGH: The image was 100px in all breakpoints before but I've now changed it to be 50px on smaller screens and 100px on larger screens.  This whole section needs some work though as there are existing layout issues that I haven't touched for now.

edwh:
Lovely to see some crossreviewing happening, since you two are much 
better able to review each other's CSS than I am.

Agree not to fix too much creep stuff, but if appropriate please add to 
the shrapnel issue.

Where will the profile shrink in this fix?  There are some cases (like 
head of a chitchat thread) where it shouldn't be too small.


JayyajGH:
Yeah it's always good to have everything reviewed.  It makes you question your solution more thoroughly. 

The only size I changed from what it what originally is at the top of the settings page.  See what you think...

edwh:
I'll have a look.  Ready to merge?

edwh:
I think it looks a bit small on mobile now, there.


From: "Edward Hibbert" <edward@ehibbert.org.uk>
To: "Freegle/iznik-nuxt" 
<reply@reply.github.com>
Sent: 06/01/2020 11:40:05
Subject: Re[2]: [Freegle/iznik-nuxt] Enhancements/use profile image 
(#157)

JayyajGH:
I'll bump it up to 75px then... And then it'll be ready to merge.

JayyajGH:
@edwh - It's now slighter bigger on mobile

JayyajGH:
I've added all of Aaron's comments to the shrapnel issue

PR #160, opened 2020-01-08
Another round of profile picture refactoring

JayyajGH:
It sometimes feels like this will never end!  Here is a bunch more.  I think there should now be about 5 or so left that I will sort out next week.


edwh:
Thanks.  In a recent checkin I've switched from the async import pattern 
to traditional import for profiles, as the async import seems to result 
in worse vue performance - I think it renders the parent component once 
the promise has been started and then has to redo it when the promise 
resolves.  Can you check your imports to this, please?  Search and 
replace will do it.

JayyajGH:
ok yup will do.  I'm about to head off now until next week so I'll look at it next Wednesday.  Feel free to update it yourself if you are bored though. :-)

edwh:
I'll probably be putting the site live before you're back.  If there's 
one thing I won't have, it's time to be bored...

JayyajGH:
Hi @edwh - Did you want the profile-image component to be imported in the standard way everywhere or just where there were performance issues?

edwh:
Everywhere, please.  It's so widely used...

JayyajGH:
OK I've now updated all of the profile-image component imports so this should now be good to go.

JayyajGH:
I've just updated the NewsLovesModal as well to use the profile image component.  There appears to be an existing copy-paste issue here as the "click=showInfo" logic on the profile image component isn't fully implemented or working.  My change leaves it as it was.  If this needs implementing then I'll add in a shrapnel TODO change to the list.

edwh:
The profile on the loves Modal shouldn't try to open another modal - too 
hard to then go back, and I don't want stacked modals.  It should just 
be a link to /profile/userid, the profile page for that user.

Can't check that right now, but if you do that, great, if not then bung 
it in an issue somewhere.  It is shrapnel.

JayyajGH:
I think I'll add it as shrapnel as it's not that straightforward and I don't want to confuse this issue.

At the moment the inner component already has click to view the profile information and this one wraps it so it makes the click events confusing and they are then tied together if you want the same logic.  You can currently click the location information to display the profile as well.  It could be that we remove it from the location and move the click handling up a level.  I'll scribble some notes and add it to an issue.

JayyajGH:
I've just checked in a final change for this.  I've now removed all of the profile styling classes from global.  One of them is still used in mydata.vue but that one fits in with the next stage of the work which is the square icons used here and also in the map components.  I'll look into them next but I'll do that under a separate PR.

edwh:
Phew!

Can you remerge from master?  There are some differences here where I have
replaced .fromNow() with | timeago, which are showing up as diffs and I
think would be lost if I merged as is.

JayyajGH:
Merged. :-)  I use GitKraken and it seemed happy to automerge.  I've given it a quick test and the profile pic stuff seems ok.  You might want to double check that you are happy with the merge.  I've checked a couple of files and they are now using timeago so they superficially look ok.

edwh:
Thanks.  This really makes me wish I'd used a component to start with.

JayyajGH:
That's easy to say in retrospect.  This kinda thing just sneaks up on you on a big project.

edwh:
Yes.  But in future I'll err on the same of creating components more 
often, I think.

Issue #116, opened 2019-12-30
Use vuelidate throughout


edwh:
- [ ] StartEndCollection:  end date > start date, no stupidly long events (3 days?), no overlapping dates, only one blank slot.
- [ ] ChatReportModal: replace if with validation
- [ ] MessageReportModal: replace if with validation
- [ ] PosterModal: replace if with validation.  name & position required, description not.
- [ ] StoriesAddModal - headline & story required

edwh:
I've added an item for photo uploading - I thought I had that somewhere but can't find it in another issue.

edwh:
Added an item for the login modal, found via Sentry.

edwh:
> I've added an item for photo uploading - I thought I had that somewhere but can't find it in another issue.

...because it's been implemented, perhaps by me.  It was a bit bugged, though, so fix shortly.

PR #164, opened 2020-01-09
Start end collection validation

nicksellen:
This validates StartEndCollection usages, with two approaches:
1. making it so you can't select invalid values in the first place
2. regular vuelidate validations that will show an error after you first submit

From https://github.com/Freegle/iznik-nuxt/issues/116 :
- [x] **end date > start date** - restricted so that it _cannot_ happen (in theory)
- [ ] **no stupidly long events (3 days?)** - Q for @edwh : this doesn't seem very long to me, what about a week long festival? do they never get added?
- [ ] **no overlapping dates** - can probably ~~steal~~ use [this code](https://gist.github.com/derickbailey/63241c2064471b9d5b6a), but I think this is a good time to introduce testing for edge cases
- [ ] **only one blank slot** - not really validation, should probably just disable the "Add another date" button if there is already a blank slot
- [x] **start times at 9am** - not really validation, just a default
- [ ] **both start and end date are provided** - I _assume_ this is needed @edwh ?


edwh:
Thanks. Will look tomorrow. On the stuff below:
- We don't get week-long things, I don't think. If we do we'll change it.
- Not sure what you're proposing to do in the overlapping case, then?
- Agree on blank slot, no action.
- 9am is a default, yes.
- Yes, we need both start and end.

nicksellen:
> We don't get week-long things, I don't think. If we do we'll change it.

We won't get any if we prevent the creation of them :p

> Not sure what you're proposing to do in the overlapping case, then?

Add testing framework. Would propose jest. For this purpose, just testing a pure function, not all the UI stuff.

edwh:
> We won't get any if we prevent the creation of them :p

We'll get someone asking why not, though, and we've had experience of having quite a lot of events which were accidentally made longer than they actually were.  So stick with this please.

> Add testing framework. Would propose jest. For this purpose, just testing a pure function, not all the UI stuff.

Yes, this kind of stuff is crying out for some test cases.  Go ahead.

PR #165, opened 2020-01-10
Upgrade LazyYoutubeVideo and use esm version

nicksellen:
Uses latest version of lazy youtube lib, updating usage to fit new way.

Also uses esm version + transpile setting to ensure es6 doesn't sneak into the output builds.

I added `ios_saf >= 9` in our browserslist config in a vain attempt to support it, it's not enough (get  `Typeerror: Type error` in the console) but maybe helps a bit...

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -309,7 +309,7 @@ module.exports = {
   build: {
     // analyze: true,
 
-    transpile: [/^vue2-google-maps($|\/)/],
+    transpile: [/^vue2-google-maps($|\/)/, 'vue-lazy-youtube-video'],
 
     extend(config, ctx) {
       if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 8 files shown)


edwh:
That looks harmless.

Did you try setting ios_saf lower?

Any ideads how to debug the type error?

nicksellen:
> Did you try setting ios_saf lower?

Yes, I put it at 8. But didn't make it work better on any of the browserstack iOS devices I tried it on. Currently it works down to the iPhone 7 / iOS 10 / Safari, below that it worked on iPhone 6S Plus / iOS9 / Chrome (but not Safari, that's the type error one).

> Any ideads how to debug the type error?

Not right now :/

edwh:
Ok, so I should merge this as is because it it does no harm.  I don't 
think we should spend further time on this just yet, though - validation 
stuff probably more important.

nicksellen:
Yes, merge away! More than do no harm, it improves the current situation - no arrow functions in the output...

edwh:
Ok.  Given that we now have two dependencies which we have had to 
transpile to fix problems, I'm wondering if we should/could tell it to 
compile all of them.  Is that feasible/wise?

nicksellen:
Not wise, will slow the build down for no benefit. Case-by-case basis.

PR #167, opened 2020-01-15
Fixed Design Issues - on find/search Page

Shivanjali1296:
Design Fixes for the find/search page mentioned for Rahul are done.
- Border around the search options.
- Focus on the input and the button

### DIFF
--- a/components/Autocomplete.vue
+++ b/components/Autocomplete.vue
@@ -1,6 +1,6 @@
 <template>
   <div :class="`${getClassName('wrapper')} autocomplete-wrapper`">
-    <b-input-group>
+    <b-input-group ref="ref_div" tabindex="0" class="ref-div">
       <input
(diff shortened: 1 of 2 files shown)


JayyajGH:
Was this PR meant to be merged rather than closed?

JayyajGH:
In case it was meant to be merged then a couple of comments:

- All colours should be defined in the _colour_vars.scss file and referenced from here.  Can existing colours be used rather than creating new ones?
- Does box-shadow need the webkit prefix?  I believe this is now supported everywhere.
- Does the ref-div:focus class need to be so specific with the extra div? 
- If activeClass always returns "search_op" then why not just list it on the element itself?
- Can the class name better reflect what it does rather than "ref-div".  That can confuse it with the name of the reference.
- Where is focus-class used?

PR #168, opened 2020-01-15
Fixed Design Issues - on find/search Page

YugaaySolutions:
Design Fixes for the find/search page.

Border around the search options.
Focus on the input and the button

### DIFF
--- a/components/Autocomplete.vue
+++ b/components/Autocomplete.vue
@@ -1,6 +1,6 @@
 <template>
   <div :class="`${getClassName('wrapper')} autocomplete-wrapper`">
-    <b-input-group>
+    <b-input-group  ref="ref_div" tabindex="0" class="ref-div">
       <input
         :id="id"
         ref="input"
@@ -70,10 +70,15 @@
 
 /* iteminp class is passed into this component in a prop */
 .iteminp ul {
+  /* 
   width: 100% !important;
   right: 0px !important;
-  padding-right: 15px !important;
-  padding-left: 15px !important;
+  margin-right: 15px !important;
+  margin-left: 15px !important;
+  */
+  /*Width adjusted for the border of the search options*/
+  width: 96% !important;
+  right: 13px !important;
 }
 
 /* postcodelist class is passed into this component in a prop */
@@ -157,6 +162,28 @@
 .autocomplete ul li.focus-list a span {
   /*backwards compat*/
   color: white;
+}
+
+
+/*Style for border round the search options*/
+.search_op {
+  border: 1px solid #61ae25;
+  border-radius: 0.2rem;
+}
+
+div .ref-div:focus {
+  background-color: #fff;
+  border-color: #80bdff;
+  outline: 0;
+  border-radius: 0.2rem;
+  box-shadow: 0 0 0 0.2rem rgba(0, 123, 255, 0.25);
+}
+
+.focus-class:focus {
+  border-color: #ced4da;
+  border-right: none;
+  -webkit-box-shadow: none;
+  box-shadow: none;
 }
 
 /*.showAll-transition{
@@ -479,8 +506,8 @@ export default {
       this.focusList = i
     },
 
-    activeClass(i) {
-      const focusClass = i === this.focusList ? 'focus-list' : ''
+     activeClass(i) {
+      const focusClass = i === this.focusList ? 'focus-list search_op' : ' search_op'
       return `${focusClass}`
     },
 
--- a/components/Autocomplete.vue
+++ b/components/Autocomplete.vue
@@ -1,15 +1,15 @@
 <template>
   <div :class="`${getClassName('wrapper')} autocomplete-wrapper`">
-    <b-input-group  ref="ref_div" tabindex="0" class="ref-div">
+    <b-input-group ref="search_div" tabindex="0" class="search-div">
       <input
         :id="id"
         ref="input"
         v-model="type"
         type="text"
-        :class="`${getClassName('input')} autocomplete-input`"
+        :class="`${getClassName('input')} autocomplete-input no-focus`"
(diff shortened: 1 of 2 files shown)


JayyajGH:
I'll move my comments across from the closed PR to this one...

- All colours should be defined in the _colour_vars.scss file and referenced from here. Can existing colours be used rather than creating new ones?
- Does box-shadow need the webkit prefix? I believe this is now supported everywhere.
- Does the ref-div:focus class need to be so specific with the extra div?
- If activeClass always returns "search_op" then why not just list it on the element itself?
- Can the class name better reflect what it does rather than "ref-div". That can confuse it with the name of the reference.
- Where is focus-class used?

components/Autocomplete.vue
```diff
@@ -70,10 +70,15 @@
 
 /* iteminp class is passed into this component in a prop */
 .iteminp ul {
+  /*
```
aaronferris: Whats the purpose of this? Are we trying to comment this code out? If so, remove it, if not remove the /*

components/Autocomplete.vue
```diff
@@ -70,10 +70,15 @@
...
 /* iteminp class is passed into this component in a prop */
 .iteminp ul {
+  /* 
   width: 100% !important;
   right: 0px !important;
-  padding-right: 15px !important;
-  padding-left: 15px !important;
+  margin-right: 15px !important;
+  margin-left: 15px !important;
+  */
```
aaronferris: Same as above, this is just noise

components/Autocomplete.vue
```diff
@@ -70,10 +70,15 @@
 
 /* iteminp class is passed into this component in a prop */
 .iteminp ul {
+  /* 
   width: 100% !important;
   right: 0px !important;
-  padding-right: 15px !important;
-  padding-left: 15px !important;
+  margin-right: 15px !important;
```
aaronferris: Can we avoid important?

components/Autocomplete.vue
```diff
@@ -70,10 +70,15 @@
...
+  margin-left: 15px !important;
+  */
+  /*Width adjusted for the border of the search options*/
```
aaronferris: Lets try and keep comments consistent, IE space after the first * and before the last *

components/Autocomplete.vue
```diff
@@ -70,10 +70,15 @@
...
+  width: 96% !important;
+  right: 13px !important;
```
aaronferris: Can we avoid the important?

components/Autocomplete.vue
```diff
@@ -70,10 +70,15 @@
...
+  /* 
   width: 100% !important;
   right: 0px !important;
-  padding-right: 15px !important;
-  padding-left: 15px !important;
+  margin-right: 15px !important;
+  margin-left: 15px !important;
+  */
+  /*Width adjusted for the border of the search options*/
+  width: 96% !important;
```
aaronferris: Can we avoid the important?

components/Autocomplete.vue
```diff
@@ -159,6 +164,28 @@
   color: white;
 }
 
+
+/*Style for border round the search options*/
```
aaronferris: Same as above, lets keep comments consistent

components/Autocomplete.vue
```diff
@@ -159,6 +164,28 @@
...
+.search_op {
+  border: 1px solid #61ae25;
+  border-radius: 0.2rem;
+}
+
+div .ref-div:focus {
+  background-color: #fff;
```
aaronferris: Same as above re: variables

components/Autocomplete.vue
```diff
@@ -159,6 +164,28 @@
...
+  border-color: #80bdff;
```
aaronferris: Same as above re: variables

components/Autocomplete.vue
```diff
@@ -159,6 +164,28 @@
...
+  outline: 0;
+  border-radius: 0.2rem;
+  box-shadow: 0 0 0 0.2rem rgba(0, 123, 255, 0.25);
```
aaronferris: is this box shadow colour re-usable? Worth a variable?

components/Autocomplete.vue
```diff
@@ -159,6 +164,28 @@
...
+}
+
+.focus-class:focus {
+  border-color: #ced4da;
```
aaronferris: Same as above re: variables

components/Autocomplete.vue
```diff
@@ -159,6 +164,28 @@
...
+  border-right: none;
+  -webkit-box-shadow: none;
```
aaronferris: Is this needed? I was under the impression autoprefixer was working out of the box with next.js, if not I would recommend we have a task to add it. Adding prefixes will become tedious.

pages/find/search/_term.vue
```diff
@@ -167,7 +167,8 @@ export default {
     this.$store.dispatch('messages/clear')
 
     // Focus on field to grab their attention.
-    this.$refs.autocomplete.$refs.input.focus()
+    //this.$refs.autocomplete.$refs.input.focus()
```
aaronferris: Remove commented code, if we don't need it it's just noise.

components/Autocomplete.vue
```diff
@@ -479,8 +506,8 @@ export default {
       this.focusList = i
     },
 
-    activeClass(i) {
-      const focusClass = i === this.focusList ? 'focus-list' : ''
+     activeClass(i) {
```
aaronferris: Odd indentation

aaronferris:
@YugaaySolutions  I have added some comments inline in the diff, apart from that welcome to the project and thanks for your contribution :)

Issue #142, opened 2019-12-31
Design shrapnel


edwh:
Little bits and pieces which could be slightly prettier, all fairly quick.

- [ ] MyData Style download as a button. We need to have it as a link so that we can use the download attribute.
- [ ] About - use a bootstrap container rather than spacer columns.
- [ ] About - Rework to use the FAQ accordion as per help page
- [ ] Disclaimer - use bootstrap container
- [ ] Terms - use bootstrap container
- [ ] Privacy - use bootstrap container
- [ ] error-toast - needs love
- [ ] DonationThermometer - animate change in value
- [ ] DraggableMap used from Spread - icon pin in the centre is not visible enough on the map.
- [ ] GroupHeader There's a max-height hack above to keep the description from hogging the screen.  It's not that pretty and some people may not even notice the scrollbar, but could be neater.
- [ ] GroupSelect - hard-coded max-width.
- [ ] OutcomeModal - Needs visual push to make people select a user.  How?
- [ ] OutcomeModal - "Please choose" is red but the dropdown shouldn't be.
- [ ] Ratings - Showing your own feedback in blue is not intuitive.  How can we improve that given the very limited space for this?
- [ ] default - Shrink navbar on scroll?  Hardcoded navbar heights in other places, though, search for 68px.
- [ ] default - Notifications dropdown list isn't wide enough before it's loaded.
- [ ] message/_id - when it doesn't exist, the

 message is brutal.  error variant is a very shouty red.
StartEndDate - should be fieldset within a form.

aaronferris:
With respect to the Group header, how do we want to position the buttons (join etc) on mobile?

Both Under the logo? Just contact under the logo?

![Screenshot 2020-01-13 at 09 17 14](https://user-images.githubusercontent.com/4068825/72244186-9f9ef880-35e5-11ea-8520-edb39951434b.png)

edwh:
Generally I don't want to take up too much vertical space.

So at that screen size:
The group logo looks too large and there's loads of whitespace.You could 
have a shrunken logo on the left side. Then on the right have the group 
name and buttons below.Then the tag line and everything else below it at 
full width.The white card could be full width on the screen (no flowery 
border) and have smaller padding round the contents.
I'd rather reposition things to reduce the amount of blank space than 
aim for consistent positioning across different screen sizes.  It's 
quite fiddly.

------ Original Message ------

aaronferris:
[...]

PR #172, opened 2020-01-18
#142 improve group header layout to avoid any overlap, obtuse or odd …

aaronferris:
Id highly recommend pulling this down and taking a look before merge because

A) I cant find our discussion history (I seem to recall trying to avoid too much vertical whitespace or a logo too large)
B) This was really fiddly across different break points (IE between lg and xl, also 991/992 as this is when the overall layout changes from 12 columns to 6. 992 to 1199 would be my main concern as there just isnt enough horizontal space to get the 3 elements sitting side by side, hence the buttons sitting underneath, as they do on mobile).

Ive applied what I feel is best, as far as size of each element is concerned but as above, open to feedback at specific breakpoints whether we want a reposition of buttons etc.

### DIFF
--- a/components/GroupHeader.vue
+++ b/components/GroupHeader.vue
@@ -13,21 +13,19 @@
         <p v-if="group.membercount" class="text-muted small">
           Founded {{ group.founded | dateonly }}. {{ group.membercount.toLocaleString() }} current freeglers.
           <br>
-          <span>
-            See
-            <nuxt-link :to="{ path: '/communityevents/' + group.id }">
-              community events
-            </nuxt-link>,
-            <nuxt-link :to="{ path: '/volunteerings/' + group.id }">
-              volunteer opportunities
-            </nuxt-link>,
-            <nuxt-link :to="{ path: '/stories/' + group.id }">
-              stories
-            </nuxt-link>, or
-            <nuxt-link :to="{ path: '/stats/' + group.nameshort }">
-              stats
-            </nuxt-link>
-          </span>
+          See
+          <nuxt-link :to="{ path: '/communityevents/' + group.id }">
+            community events
+          </nuxt-link>,
+          <nuxt-link :to="{ path: '/volunteerings/' + group.id }">
+            volunteer opportunities
+          </nuxt-link>,
+          <nuxt-link :to="{ path: '/stories/' + group.id }">
+            stories
+          </nuxt-link>, or
+          <nuxt-link :to="{ path: '/stats/' + group.nameshort }">
+            stats
+          </nuxt-link>
         </p>
       </b-col>
       <b-col cols="12" md="4" lg="12" class="group-header-buttons">
--- a/components/GroupHeader.vue
+++ b/components/GroupHeader.vue
@@ -1,10 +1,10 @@
 <template>
   <b-card bg-light>
     <b-row v-if="group.profile" class="mt-1">
-      <b-col cols="4" sm="2" lg="3" xl="2">
+      <b-col cols="4" md="2" lg="3" xl="2">
         <b-img-lazy rounded thumbnail alt="Community profile picture" :src="group.profile" class="js-pageimage" />
       </b-col>
-      <b-col cols="8" sm="5" lg="9" xl="5">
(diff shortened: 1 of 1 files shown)


edwh:
Thank you.  This needs a careful look, so I won't rush it.

edwh:
It certainly looks better.

There are still a few niggles.  In this example the Join/Leave button doesn't show because I am a moderator on that group.  But I think the effect is the same.

At around 500 width it looks like this:

image

I think if we could put the "See community events" across the whole width it would save a line break.

At around 600, it looks like this:

image

The image is small and the text squished - I think it would look better as it does at the 500ish display there, where the buttons move onto the next line.

At the widest:

image

I think the contact button is taking up too much space here - the image could be slightly bigger (but it's ok), and if the central section had more space it wouldn't line break.
You get similar effects when you keep growing and the sidebars kick in and shrink the central space.

aaronferris:
We need to be careful that we dont start adding specific breakpoints for various use cases, Bootstrap uses 4 and we should try and stick to these for consistency, I would recommend

Point 1

This is a tricky one, if we look at moving this text it starts to look disconnected from the group description at say 360.

Screenshot 2020-01-19 at 10 14 48

Screenshot 2020-01-19 at 10 15 34

Point 2

I would recommend we position the buttons underneath for all devices < 768 as there isnt enough real estate to handle the buttons top right. Same for 992-1199, same problem.

600
Screenshot 2020-01-19 at 10 58 39

768
Screenshot 2020-01-19 at 10 58 29

This would also be the case for 992-1199
Screenshot 2020-01-19 at 10 59 35

Point 3

This is a tricky one, because we are sticking with bootstrap grids anything less than 5 columns for the text means there isnt enough space for the 'contact volunteer' button at the XL breakpoint (1200)

Screenshot 2020-01-19 at 10 11 33

Ive tried to remedy this by overriding the bootstrap column widths, and the gutters for > 1200. The screenshot in your example is missing the join button which does = more horizontal whitespace.

Screenshot 2020-01-19 at 11 02 27

Screenshot 2020-01-19 at 11 02 34


If this still doesn't resolve and do what we want, a rewrite of this to not use bootstrap grids would be on the cards.

edwh:
Definitely agree with sticking to Bootstrap breakpoints.

On point 1, I think the text would look less disconnected at 360 if the image was larger.  Not sure whether that's feasible, though - the tagline on the right is variable length for different groups.  But I'm ok with it looking as it does in your screenshot.

Point 2 - ok, underneath it goes.

Point 3 - this looks ok.  Wasted whitespace is less of an issue on larger screens, so this looks ok to me.

Ready to merge, then?



The links are causing spaces before commas, which is mildly annoying - might be my IDE formatting.

aaronferris:
So I think we're good with point 2 and 3, so just point 1 needs to be remedied before merge

Ive increased the image size an extra column, doesn't look too bad

Screenshot 2020-01-19 at 11 19 53 (2)
Screenshot 2020-01-19 at 11 19 57

Re: the commas, that's because they exist outside of the link creation, we could include them as part of the link text instead. I dont confess to be an SEO expert but I think we're safe including punctuation in link text, providing it doesn't form part of the URL to cause any crawler confusion.

`      Founded {{ group.founded | dateonly }}. {{ group.membercount.toLocaleString() }} current freeglers.
          <br>
          See
          <nuxt-link :to="{ path: '/communityevents/' + group.id }">
            community events
          </nuxt-link>,
          <nuxt-link :to="{ path: '/volunteerings/' + group.id }">
            volunteer opportunities
          </nuxt-link>,
          <nuxt-link :to="{ path: '/stories/' + group.id }">
            stories
          </nuxt-link>, or
          <nuxt-link :to="{ path: '/stats/' + group.nameshort }">
            stats
          </nuxt-link>`

I would assume its just a case of including them as part of the link text as so

edwh:
Ok, I'll merge as is.

I kind of get irked by the comma being coloured as part of the link, 
though.  But there are more important things just now...

edwh:
Merged.  Thanks for fiddling with this.  It's much better now.

PR #173, opened 2020-01-19
#142 potential header changes with the 'see' text - do not merge

aaronferris:
This isnt for merge, its in case we wanted to go with discussions from point 1 from 

https://github.com/Freegle/iznik-nuxt/pull/172


edwh:
Looks better, thanks.  Only issue is upto 765, where there's room to put it on the right hand side rather than below.
image

Issue #124, opened 2019-12-31
Accessibility


edwh:
Various accessibility issues.  We've not yet done a proper accessibility review; this issue is a placeholder for specific things we already know.

- [ ] The availability modal isn't accessible.  Note that this may change because of #123.
- [ ] We should review all <a> and <nuxt-link> to see if they require aria-label.
- [ ] We should check if we have any image links without alt text.  Can we lint this?

PR #176, opened 2020-01-22
Enhancements/add profile image alt text

JayyajGH:
I'd accidentally removed the specific alt text in these places before so this is adding it back in. It is definitely required in these places as it isn't specified elsewhere.

I'm still thinking about what to do with the others.  It might be that alttext is left blank in the other places but I need to check if the username is provided in some other way.  At the moment it just says "Profile picture" as it did before.

I've also removed "title" from the component.  It is generally accepted that title is bad as it can interfere with alttext on some screenreader software.  If altext is present (even if it is blank) it will get used over title.


edwh:
I'll merge this one.

Based on what you said about title, I googled and found 
https://silktide.com/blog/2013/i-thought-title-text-improved-accessibility-i-was-wrong. 
Gosh.  So, shall we have an issue to remove title attributes everywhere?

JayyajGH:
Things like that can be done when we review the accessibility in general.  As this was a new component then I thought it should be correct from the state.  I think these are generally just opinions.  There is no black and white. Alttext is the important one that is the most widely supported so as long as that one is correct that is the main thing.  I think if you have title as well (especially if they aren't the same - including capitalisation and punctuation) then they can be read twice by screenreaders.

edwh:
I'll add that link into the #124 .

PR #178, opened 2020-01-23
Change the login elements to have flex-basis set as auto rather than …

JayyajGH:
To steal a comment from a random site:

"Since you're setting the flex direction to column, the flex-basis value goes in the vertical direction, but the height of the container is determined by its children, so you have a circular reference and a percentage value is meaningless.

However, the items should, by default, not shrink below their minimum content size, but in Safari they're not honoring that. When you say 100%, since the height isn't fixed, that computes to 0, which is why you're seeing them all squished."

So basically 100% doesn't really make sense for the height but it's supported everywhere apart from Safari 9 and IE11.

I've only tested this very briefly at the moment.  I've tried it on Chrome and Firefox on desktop and using devtools on Safari 9.  Looks like the problem also happens on IE11 too which I've also only tested using devtools.

### DIFF
--- a/components/LoginModal.vue
+++ b/components/LoginModal.vue
@@ -529,7 +529,7 @@ $color-google: #4285f4;
 $color-yahoo: #6b0094;
 
 .signin__section--social {
-  flex: 0 1 100%;
+  flex: 0 1 auto;
 
   @include media-breakpoint-up(lg) {
     flex: 0 1 37%;
@@ -537,7 +537,7 @@ $color-yahoo: #6b0094;
 }
 
 .signin__section--freegle {
-  flex: 0 1 100%;
+  flex: 0 1 auto;
 
   @include media-breakpoint-up(lg) {
     flex: 0 1 44%;


edwh:
If you go to the screenshots section on BS, you could put in an URL for 
a page which popped up the Login modal, and then get screenshots on lots 
of devices to see if it worked?

JayyajGH:
I've not used the screenshots option before.  Sounds good.

I only made my change in devtools so it won't work on a URL unless I get the local build running which I've not tried before. And because it's a popup modal there isn't a unique URL to get it to display is there?

edwh:
If you go to http://localhost:3000/chitchat , for example, it will pop 
up a modal because you need to be logged in to see that.

JayyajGH:
Ah right. Yes! I'll try the build option then.  Got a lot of meetings today so will be dipping in and out of freegle.

JayyajGH:
Ok I've tried a bunch of screenshots and they all look ok to me.

edwh:
Ok, thanks.  It's still marked as a WIP, but once it's ready I'll merge 
it.

JayyajGH:
Just as a caveat...  I've only tested this in devtools and in various browsers on browserstack.  I haven't been able to test it in an actual real-life browser yet.  I would feel more comfortable if we did that first.  I can do this tomorrow morning if you don't have time this afternoon.

edwh:
I can test on real browsers.

PR #179, opened 2020-01-23
Use flex and margins to make the chat messages more flexible for long…

aaronferris:
… messages

### DIFF
--- a/components/ChatMessageText.vue
+++ b/components/ChatMessageText.vue
@@ -56,22 +56,25 @@
   .chatMessage {
     margin-left: auto;
   }
+
+  .chatMessageProfilePic {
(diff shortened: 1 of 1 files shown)


edwh:
Looks good, including on IE11.  Ta.

JayyajGH:
@aaronferris @edwh 

A bit of a belated message on a closed PR but can the IE11 hack now be removed from this file?  The setting of the min-width on the media class.

edwh:
I'll test that, but it'll be next week.

edwh:
Ah, yes, that CSS is referencing something which no longer exists.  So probably it gets shaken out but I'll remove it.

PR #183, opened 2020-01-27
Replace hardcoded bootstrap breakpoints with bootstrap functions

JayyajGH:
Remove the hardcoded breakpoints and use the bootstrap breakpoint functions instead.  They didn't actually correspond to the bootstrap ones anyway!

As far as I can tell you override the bootstrap versions by defining $grid-breakpoints.  When you hardcoded the values were you attempting to redefine the bootstrap ones or just have something to use in the code?  Just checking that I'm not missing anything.

### DIFF
--- a/pages/stats/_groupname.vue
+++ b/pages/stats/_groupname.vue
@@ -149,43 +149,7 @@
     </b-row>
   </div>
 </template>
-<style scoped lang="scss">
-@import 'color-vars';
 
-$bootstrap-sm: 768px;
-
-.chart-wrapper {
-  flex-direction: column;
-
-  @media (min-width: $bootstrap-sm) {
-    flex-direction: row;
-  }
-}
-
-.card {
-  &.chart {
-    height: 100%;
-  }
-}
-
-.titlelogo {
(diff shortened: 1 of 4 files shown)


pages/stats/_groupname.vue
```diff
@@ -480,3 +444,42 @@ export default {
...
+.purple {
+  color: $color-purple !important;
+}
+
+.gold {
+  color: $color-gold !important;
+}
+
+.green {
+  color: green !important;
```
aaronferris: Appreciate this is legacy but should really be a variable.

edwh: True.  I'll fix that after merging.

JayyajGH:
@aaronferris  - I'll do another PR for the colour change...

pages/stats/_groupname.vue
JayyajGH: I thought I'd got all of the colours when I created the variable file but obviously not!

JayyajGH:
Unless @edwh gets there first.  I'm done devving for the day now though.

PR #185, opened 2020-01-28
Move all remaining colour variables into the colour file.

JayyajGH:
Move a bunch more colours into the colour file.  I've not done any consolidation as part of this.  I've just moved their definitions into the common location.  It might be able time for us to see if we can chop any out.

### DIFF
--- a/assets/css/_color-vars.scss
+++ b/assets/css/_color-vars.scss
@@ -22,6 +22,8 @@ $color-gray--base: #A9A9A9;
 $color-gray--dark: #808080;
 $color-gray-1: rgba(221,142,98,0.05);
 $color-gray-2: rgba(62,108,178,0.02);
+$color-gray-3: #E9ECEF;
+$color-gray-4: #CED4DA;
 
 $color-blue--x-light: #D1ECF1;
 $color-blue-x-light2: #BEE5EB;
@@ -30,6 +32,7 @@ $color-blue--light: #4895DD;
 $color-blue--base: #0000FF;
 $color-blue--dark: #35495E;
 $color-blue--1: #0C5460;
+$color-blue--2: #204D74;
 
 $color-yellow--light: #F9F7EC;
 $color-yellow--base: #F3F0DA;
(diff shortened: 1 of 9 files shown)


edwh:
There's certainly an argument that we have too many, isn't there?

JayyajGH:
We do have quiet a lot! I think some consolidation would help to bring together the Freegle brand a bit more. Although in reality that's probably just marketing spiel and no one would actually notice. :-)

Probably the main advantage to us as developers and for developers new to the project, the benefit would be the simplicity of adding new elements.  What colour should I make this background?  Grey 1, 2, 3, 4 or 5?  And which one of the 8 blues should the foreground be?

It doesn't really matter whether the colours are coming from bootstrap or our own custom ones.  If we are using them then they are part of the project and therefore brand.  We're talking icing on the cake here I think but it's definitely worth doing at some point.

aaronferris:
I think if we compared the colours, we'd find some are very similar to the naked eye so could be combined into a single variable. That's an assumption based on experience of a large number of variables.

Having a lot isn't a problem per se, so long as we're consistent in our usage of them.

IE: Links are colour $green, buttons are colour $red, <p> tags are $dark-grey etc. it becomes a problem when we start mixing and matching elements with different colours when their primary purpose is the same. IE links being a different colour one one page compared to the next page. More colour options = more likelihood of this happening.

It'll be time consuming as we need to be careful of regression but agreed with @JayyajGH it's worth doing at some stage.

edwh:
We'd probably just have to start, have a look at some uses of colours, 
and build up a definition of what each is for.

Not sure when any of us will feel like that's the funnest thing to do 
today, though.

aaronferris:
No its not a 4:10pm job for sure, its a 'fresh set of eyes 9am with a coffee' job.

We might also want to consider implementing something like pattern labs in the future. 

https://github.com/pattern-lab/patternlab-php

It's basically a design guide, but you can use it for atomic designs, colours, layouts all sorts. Makes it very easy to see what images, colours etc are used in the site. That's a massive 'nice to have' though. Ive used it on projects in the past and found it very useful.

PR #187, opened 2020-01-29
Don't fix the height of the donate message

JayyajGH:
Here is an option for fixing this issue.  If this is acceptable then I'll confirm the PR.

### DIFF
--- a/components/JobsTopBar.vue
+++ b/components/JobsTopBar.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="location" class="mb-2 jobbox bg-light overflow-hidden">
+  <div v-if="location" class="mb-2 bg-light">
     <NoticeMessage v-if="blocked" variant="warning">
       <h3>Please help keep Freegle running</h3>
       <p>
@@ -13,7 +13,7 @@
         </b-btn>
       </a>
     </NoticeMessage>
-    <div v-else>
+    <div v-else class="jobbox overflow-hidden">
       <div class="mb-1 text-center small text-muted">
         Jobs near you.  Freegle gets a small amount if you click, which helps keep us going.  <nuxt-link to="/jobs">
           See more jobs
@@ -25,11 +25,7 @@
     </div>
   </div>
 </template>
-<style scoped>
-.jobbox {
-  height: 220px;
-}
-</style>
+
 <script>
 import Job from './Job'
 import NoticeMessage from './NoticeMessage'
@@ -77,3 +73,9 @@ export default {
   methods: {}
 }
 </script>
+
+<style scoped>
+.jobbox {
+  height: 220px;
+}
+</style>


aaronferris:
Code looks OK but I'm always apprehensive about set fixed heights, makes the content very inflexible, especially when someone starts playing with the font size of their browser.

JayyajGH:
Yeah it can look a bit odd.  With overflow: hidden it cuts off the text halfway through as well.  I've not touched the job listings though so they work as they did before - with a fixed height.  I've just removed the fixed height from the donate message so the whole thing shows to fix the issue listed on Slack.

edwh:
Just removing the jobbox CSS entirely seems to work ok for me.  I think I had some idea that the top box of jobs would be scrollable, but it doesn't seem necessarily.  Does that look ok to you?

JayyajGH:
I'm guessing if there are a lot of jobs then it would take up too much of the screen though.  How many jobs could the API return?  Maybe we bin the max-height and put a max on the number to display of 3 for example.

edwh:
It will display 5 here by default, so it's bounded.

JayyajGH:
At the moment it looks like 3 will fit in the space allocated.  This takes up about 1/2 of the screen.  Would 5 be too much?  Should we restrict it to 3 and remove the max-height?

edwh:
Or maybe add classes to the bottom two to hide them on small screens?

edwh:
Something like this seems to work ok, and the same for recently viewed.

```
      <div v-for="(job, index) in jobs" :key="'job-' + job.onmousedown" class="">
        <Job :summary="true" :job="job" :class="index > 1 ? 'd-none d-md-block' : ''" />
      </div>
```

I'll push that, I think.  Fixes the donate button and also is less of a screen hog on mobile.

JayyajGH:
Ok cool.

JayyajGH:
Shall I just delete this PR?

edwh:
Yeah, I'll push the fix.  Thanks, though, it's doing this PR that got a 
fix, even if it's a different fix.

JayyajGH:
A belated comment on this @edwh 

What is the Freegle traffic split on desktop/tablet/mobile?  By doing this are we chopping out a load of revenue from mobile?

edwh:
It's about 50/50.  The purpose of the box at the top is not so much 
because the specific jobs we choose to show there will be ones people 
click on (that's unlikely).  It's to show them that there are jobs and 
encourage them to click to see more of them.

  But realistically most of our job ad revenue is likely to come from 
larger devices than small phones.

JayyajGH:
As an alternative solution, on small devices we could show a box with a link to jobs instead of showing the jobs themselves.

edwh:
We could, but I think specific examples are what would prompt people to 
click.  Otherwise half of them will misinterpret it as "Jobs with 
Freegle" and ignore it.

....but classic case of something to actually trial once I have some 
bandit testing infrastructure.


JayyajGH:
That makes sense.  Yeah definitely a perfect example of an A/B test requirement.

PR #188, opened 2020-01-30
Create a donation button and use it where required.

JayyajGH:
Here is a new donation button component.

Things to note:
- I've moved the paypal logo locally in case it disappears from the external site.
- I've added "noopener noreferrer" to the link for security reasons.
- I've made it accessible so a screenreader will just read out "Donate to Freegle with paypal (opens in a new tab)".  Although this worked before it was slightly convoluted as it was reading out the image alt text and then the text of the link. This way we are in complete control and it means I can also add that it opens in a new tab.

I haven't tested this everywhere yet which is why I've just created it as a draft at the moment.


JayyajGH:
Should the donate button on this page under "text alerts" look like the standard donate button used elsewhere?
https://www.ilovefreegle.org/settings

edwh:
Yes, it should.  Thanks.

JayyajGH:
OK cool.  I'll update it...

JayyajGH:
This PR is now ready

edwh:
Thanks, that's needed.  There will be a slight textual change on the Settings page, but that's ok.

PR #189, opened 2020-01-31
Refactor some of the more complex logic in the HTML into computed props.

JayyajGH:
Some of the logic in the HTML in this component was complex and duplicated so I have refactored it out into computed props.  It makes the HTML much nicer to read now too.

### DIFF
--- a/components/ChatMessageText.vue
+++ b/components/ChatMessageText.vue
@@ -46,10 +46,13 @@ export default {
   extends: ChatBase,
   computed: {
     messageIsFromCurrentUser() {
-      return this.chatmessage.userid === $store.state.auth.user.id
+      return this.chatmessage.userid === this.$store.state.auth.user.id
     },
     messageIsNew() {
-      return chatmessage.secondsago < 60 || chatmessage.id > chat.lastmsgseen
+      return (
+        this.chatmessage.secondsago < 60 ||
+        this.chatmessage.id > this.chat.lastmsgseen
+      )
     }
   }
 }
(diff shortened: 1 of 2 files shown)


edwh:
Using computed properties is neater, thanks.

There's "myChatMessage" and "chatMessage theirs" vs "chatMessage mine".  It feels weird to have both of these - shouldn't we be consistent on whether we have a single class or multiple?

For the bits like this:

```
          <div class="chatMessageProfilePic">
            <div>
              <profile-image :image="me.profile.turl" class="ml-1 mb-1 mt-1 inline" is-thumbnail size="sm" />
            </div>
          </div>
```
...can't we lose some divs?  I can't see a good reason for some of the spans, either. 

(This is my mess, obvs.)

JayyajGH:
Might be easiest if @aaronferris answers as he did the last round of refactoring here. I could just remove the extra div and rename the classes but I know this was a complex area and I didn't follow the original discussion thread that closely and I don't want to mess it up without fully understanding the issues.

It's also been in my head to refactor this section to not have the duplication of the two blocks.  i.e. just one block for profile pic and message and flip the order around with flexbox for yours/mine and then drive an overall class from the computed prop.

Maybe merge this PR and talk about a refactor independently?

JayyajGH:
I've just knocked something up very roughly to show what I mean.  This isn't complete or indeed tested but it superficially does what I mean.  I'll create a draft PR...

edwh:
> Maybe merge this PR and talk about a refactor independently?

Yes.

JayyajGH:
There is now a draft PR to address some of the things discussed here...

PR #190, opened 2020-01-31
First pass at DRYing out the component

JayyajGH:
DRY out the component.  This is work in progress and in no way complete or tested.


edwh:
Well, that's much better.  

I didn't know about the kind of conditional class construction you're using there.  I bet it works with a space after the brace, though :-).

Have you tried this on BrowserStack to make sure it works with e.g. IE11 and Safari 9?

JayyajGH:
Yeah it looks quite tidy doesn't it.  And removes the duplication of the two blocks.

I think we need to be careful with the use of flexbox "order" as it can mess up tabbing and accessibility but in this case it doesn't matter as they aren't interactive elements.

I've so far given it a very brief test in Chrome and that's it.  If you like the refactor then I'll give it a proper test.  It shouldn't really cause any issues as it still just uses all of the code that Aaron added.  I've just driven it off a slightly different set of classes.

edwh:
I do like it.  How it should have been from the start.

JayyajGH:
I'll give it a test then.  If all the code was exactly how you wanted it from the start then you never would have got it live!

edwh:
You're not wrong about that.

JayyajGH:
OK I've now given this a test in Chrome and Safari 9 and it all looks good.  I struggled with getting freegle working at all on localhost in browserstack using IE9 but it seemed to work when I opened the devtools.  It looks fine as well. I've now marked this ready for review.

edwh:
Can you merge for conflicts?

IE9 won't work - it's IE11 that's the other odd one.

JayyajGH:
When I said IE9 I meant IE11...

JayyajGH:
@edwh Conflicts should now be resolved

edwh:
Ah, ok.

I see that same thing in IE11, where when I fire it up I don't see a sign in button, then I open the dev console and it all works.  If I try opening the dev console before loading the site to debug it, it kindly closes it for me.  I don't know if real users are seeing that.

JayyajGH:
I've just tried the live site on a real IE11 and it works fine for me.  Just not in browserstack.

edwh:
Well, that's reassuring.  I think I've seen the same thing on my 
machine, though.

PR #192, opened 2020-02-05
Enhancements/refactor duplicated code in message

JayyajGH:
The start of some work to remove the duplication of code in the Message.vue component

### DIFF
--- a/components/Message.vue
+++ b/components/Message.vue
@@ -17,15 +17,6 @@
               </span>
             </div>
             <MessageHistory :id="id" :groups="groups" />
-            <!--
-            <div v-for="group in groups" :key="'message-' + id + '-' + group.id" class="small muted">
-              <div class="small">
-                {{ group.arrival | timeago }} on <nuxt-link :to="'/explore/' + group.groupid">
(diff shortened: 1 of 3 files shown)


edwh:
I think  `:display-message-link="true"` is better as display-message-link - Nick pointed me at [this](https://vuejs.org/v2/guide/components-props.html#Passing-a-Boolean) earlier.

Does the default of the groups prop need to be a function for some reason?

JayyajGH:
:point="true". Or should I say "point". :-)  Good spot.

JayyajGH:
I lifted the groups prop definition straight from the original parent component.  Not sure why that was done originally so I just replicated it.

edwh:
I can't imagine a reason.  If that's the worst grunge in the code we can live with it.

JayyajGH:
A belated comment re the groups prop....

This comment:
 // Object or array defaults must be returned from a factory function

is from here:
https://vuejs.org/v2/guide/components-props.html

So it looks like the default of an object or array should be a function

edwh:
Aha.

PR #193, opened 2020-02-05
Sort out the anchor semantics and tidy the classes

JayyajGH:
The "Post a WANTED" button was a button within an anchor.  It should just be an anchor.  Also sort out the duplicate elements and just makes the classes conditional.


edwh:
I think that predates me bothering to read and notice that you could put `to` on buttons.  It's one of the first components I wrote, I think.

PR #196, opened 2020-02-07
Move the local storage checks into a component

JayyajGH:
Move the local storage functionality into a component to start to tidy up the default layout

### DIFF
--- a/components/LocalStorageMonitor.vue
+++ b/components/LocalStorageMonitor.vue
@@ -1,0 +1,81 @@
+<template>
+  <div v-if="localStorageErrors" class="storage w-100 text-center">
+    <b-alert show variant="danger">
+      <p>We can't access local storage on your browser.</p>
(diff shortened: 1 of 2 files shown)


edwh:
Thanks.  There's a lot to tidy up in there, I'm afraid.

There's a bug in this code - I see the warning when I don't believe it.  Possibly something to do with multiple tabs being open.

JayyajGH:
I'm guessing by the fact that you merged it, you meant the bug is an existing bug rather than something I have introduced?

Most of the file appears to be the navbar.  I've tried a few ways of tidying it but had some issues so I need to think of another angle I think!

edwh:
Yes, the bug is existing and so far undiagnosed.

Maybe see if you can combine the small and large navbar first? Not sure whether that will work but it would be an alternative approach.

JayyajGH:
Yeah that was one of my first thoughts too.  There are so many different parts to the navbar are in different orders etc. that it's not the easiest of refactors.  I'll have a play...

PR #197, opened 2020-02-10
Enhancements/add error page

JayyajGH:
It doesn't look pretty at the moment but can I get your input on wording please @edwh


edwh:
I might tweak the punctuation but it's basically fine, yes.

JayyajGH:
I'm still on the hunt for a suitable image to use to finish this off at the moment.

edwh:
With no copyright issues, of course.

JayyajGH:
Yeah that's the tricky bit.  Where did the video come from?  Can something be created with the same characters?

edwh:
Yes, it probably could, actually - it comes from a friendly firm.  
Describe what you want and I'll ask them.

JayyajGH:
If we want to stick to the cartoony style then something like a confused man or someone tearing their hair out in similar style to the video.  I'm thinking something along the lines of the Pixar 404 page.

https://www.pixar.com/hfeature-films-launch

edwh:
Yes, our friendly people will do one for us.  They are lovely.

JayyajGH:
Awesome!  It would be good to keep things in a similar style.

edwh:
Here's one from the nice video people.

freegle 404 character

JayyajGH:
Awesome.  I'll insert it into the new error page when I have a chance.

JayyajGH:
OK what about something like this...

JayyajGH:
At the moment it's the same image for all errors but with slightly different error text.

PR #198, opened 2020-02-11
Update the naming of some of the methods in the store

JayyajGH:
Whilst thinking about a refactor of the default.vue layout I've been looking at moving notifications out into a component.  Whilst doing this I've been struggling to work out what is going on and I think this might be partly due to the naming of some of the store methods.  For example there is a getter and an action both called "count".  One returns the current fetched count and the other actually gets the latest count.
Because of this I've tried to rename a few of the methods to better reflect their purpose.  So far I've just updated it in the store code and not where it is used so we can start a discussion.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -318,16 +318,18 @@ export default {
       return this.$store.getters['auth/user']
     },
     notifications() {
-      const notifications = this.$store.getters['notifications/list']
+      const notifications = this.$store.getters['notifications/getCurrentList']
 
-      notifications.sort(function(a, b) {
-        return new Date(b.timestamp) - new Date(a.timestamp)
-      })
+      if (notifications) {
+        notifications.sort(function(a, b) {
+          return new Date(b.timestamp) - new Date(a.timestamp)
+        })
+      }
 
       return notifications
     },
     notificationCount() {
-      return this.$store.getters['notifications/count']
+      return this.$store.getters['notifications/getCurrentCount']
     },
     chatCount() {
       return this.$store.getters['chats/unseenCount']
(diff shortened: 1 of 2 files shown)


JayyajGH:
On this subject, what is the purpose of the infinite loading for notifications?  When you drop down the notifications then the latest list is fetched and it is also reactive.  Or am I missing something?

edwh:
Some of the perf analysis I've done shows that it can be significantly 
better to use an infinite loader as a way of delaying the insertion of 
data into the DOM until you need it.  I think notifications was one of 
those.  Otherwise we do a lot of expensive DOM manipulation to create 
some stuff that most people will never see.

edwh:
JobsSidebar is a better example of what I just described, where we have 
it all in memory but use infinite scroll to reduce the DOM.

For notifications it looks like I did a more traditional infinite scroll 
where chunks are fetched from the server, but then broke it later when I 
tried to do a speed-up.  I've pushed a fix.  The key is the context 
field of the store, which means that list goes one further back each 
time.  I think if I were writing that now, I'd move the context logic 
out of the store rather than hide it in there.

edwh:
Meanwhile, back to your changes...the name changes look ok except for 
the rename of list to fetchLatestList - because of the use of context 
it's more like fetchNextChunk.

JayyajGH:
Ah ok that makes sense. Thanks for the explanation.  Without seeing the server code it was hard to tell. Ok I'll make a tweak to the naming...

JayyajGH:
Are you happy with this naming now?  If so then I'll implement it fully.

edwh:
Go for it.

JayyajGH:
:+1:

JayyajGH:
Ok take a look at this and see what you think.  I'm mainly doing this to assist me with the navbar refactoring but it should hopefully make things generally more readable.

JayyajGH:
Actually I still don't think this is right.  With my current profile, I have 5 notifications in the list in the store but the count in the store is 0.  Is count actually unread count?  Or how many notifications haven't yet been fetched?

edwh:
The count is unread count.

JayyajGH:
ok cool.  I'll update the name accordingly then.

PR #201, opened 2020-02-13
Remove the unnecessary alt attribute

JayyajGH:
Ok this is an interesting one....

I’ve been having a think about alt text for user uploaded images after seeing your recent checkin to modImage.vue.  Basically in your checkin, all photos will have alttext of “Message photo”.  Generally this isn’t helpful at all to someone that can’t see the image.  It doesn’t provide any useful information.  

Previously I would have suggested that these type of images have empty alt text. i.e. alt=”” but after having a further read on the subject I think I’ve changed my mind.

If the image conveys information then it should have alt text. Setting alttext to be an empty string states that the image doesn’t provide any extra useful information. e.g. just an icon on a cancel button for example.  In the case that the user has uploaded an image then the information is providing useful information but it is just not possible to provide alttext. We aren't going to get the Freeglers to add in alt text to each photo. In this case the alttext attribute should be missed out.

Here are some links that discuss this:

https://www.joedolson.com/2015/03/are-alt-attributes-required-always/

https://www.w3.org/TR/2017/REC-html52-20171214/semantics-embedded-content.html#guidance-for-markup-generators

From this link it states that “omitting this attribute indicates that the image is a key part of the content but no textual equivalent is available.  Setting this attribute to the empty string indicates that this image is not a key part of the content; non-visual browsers may omit it from rendering.”

If you add the attribute "generator-unable-to-provide-required-alt" then it can signal to conformance checkers that the image is ok.

### DIFF
--- a/components/ChatMessageCompleted.vue
+++ b/components/ChatMessageCompleted.vue
@@ -10,7 +10,7 @@
                 class="float-right"
                 rounded
                 thumbnail
-                alt="Item photo"
+                generator-unable-to-provide-required-alt=""
                 lazy
                 :src="refmsg.attachments[0].paththumb"
                 width="70px"
@@ -38,7 +38,7 @@
                 class="float-right"
                 rounded
                 thumbnail
-                alt="Item photo"
+                generator-unable-to-provide-required-alt=""
                 lazy
                 :src="refmsg.attachments[0].paththumb"
                 width="70px"
--- a/components/ChatMessageImage.vue
+++ b/components/ChatMessageImage.vue
@@ -11,7 +11,7 @@
                 rounded
                 thumbnail
                 class="chatimage"
-                alt="Chat image"
+                generator-unable-to-provide-required-alt=""
                 :src="chatmessage.image.path"
                 @error.native="brokenImage"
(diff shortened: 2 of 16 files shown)


edwh:
Interesting.  Doesn't the same apply all over the place, e.g. ChitChat, 
messages themselves, community event/volunteer op photos?

JayyajGH:
Yeah I think it probably does.  I think generic alt text for images like "image" or "Item image" doesn't help in any way at all.  I would previously have set these to be alt="" but I can see why that isn't quite right either.

JayyajGH:
Shall I update all the obvious ones? At least they are then easy to find should we change our mind or something new comes up.

JayyajGH:
Alternatively we could ask users to describe each picture they upload in detail... :-)

JayyajGH:
Done...

edwh:
If you could update the obvious ones, that would be great.

I'll send you a list of the emails of all freeglers so that you can 
start mailing them to ask them to include informative alt text.

JayyajGH:
The obvious ones are now done.

PR #206, opened 2020-02-19
Restyle the download button - From design shrapnel issue.

JayyajGH:


JayyajGH:
Removed from design shrapnel issue too.

PR #211, opened 2020-02-26
Use the profile image component

JayyajGH:
A couple more places where the profile image component should be used.  The sizing here was massive before so I've made it a little smaller.  See what you think.  The layout of the header section generally needs a bit of work on various screen sizes but this doesn't make that any worse.

### DIFF
--- a/components/ProfileModal.vue
+++ b/components/ProfileModal.vue
@@ -10,7 +10,7 @@
       <div class="m-0 coverphoto">
         <b-media>
           <template slot="aside">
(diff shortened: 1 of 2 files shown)


edwh:
It needs to be significantly larger here because there are some people 
who have very busy profile images which are hard to interpret, and the 
natural reaction is to click and see a bigger one.

I think it does look worse (sorry) because the image no longer is the 
full height of the top bar and therefore looks off-centre both 
vertically and horizontally.



JayyajGH:
Ok no worries. I'll have a play with it and see what I can do.

PR #212, opened 2020-02-27
Sort out the use of flexbox on the large navigation menu.

JayyajGH:
After investigating the IE11 issue I found a few oddities with the way the navbar was laid out.  The centre toggle with a 100% width and centre justification and the right hand nav with an auto left margin just seemed to be confusing the hell out of flexbox. I think it was probably the addition of the centre toggle that broke this in IE11.

I've just removed all the weird stuff and told the nav parent to justify the content.    I've not touched the mobile navigation.

I've tested this in IE11 on browserstack and in devtools in Chrome so this will still need a proper test on different browsers/devices.  If I don't get a chance to do this today then I'll do it tomorrow.

I thought I'd check this in as you could test it if you were desperate for a fix!

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -14,8 +14,8 @@
       </b-navbar-brand>
       <client-only>
         <b-navbar-toggle v-if="loggedIn" target="nav_collapse" />
-        <b-collapse v-if="loggedIn" id="nav_collapse" ref="nav_collapse" is-nav class="flex-nowrap">
-          <b-navbar-nav class="justify-content-between">
+        <b-collapse v-if="loggedIn" id="nav_collapse" ref="nav_collapse" is-nav class="flex-nowrap justify-content-between">
+          <b-navbar-nav>
             <b-nav-item id="menu-option-mygroups" class="text-center small p-0" to="/communities" @mousedown="maybeReload('/communities')">
               <v-icon name="users" scale="2" /><br>
               Communities
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -126,8 +126,8 @@
       </b-navbar-nav>
     </b-navbar>
     <!-- Navbar for small screens -->
-    <b-navbar id="navbar_small" toggleable="xl" type="dark" class="ourBack d-flex justify-content-end d-xl-none" fixed="top">
-      <b-navbar-brand to="/" class="p-0 mr-auto">
+    <b-navbar id="navbar_small" toggleable="xl" type="dark" class="ourBack d-flex justify-content-between d-xl-none" fixed="top">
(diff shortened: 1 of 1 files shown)


JayyajGH:
Right I've now tested this on a few browsers and it seems ok.  Are you easily able to test it on IE11?  I've only tested it on IE11 in devtools and it looks ok.

edwh:
Looks ok on Chrome, Edge, FF.  On IE11 the logo is positioned wrongly and the hamburger disappears:

image

The centering of "Communities" looks a bit off though that's an existing problem.

JayyajGH:
Hmmm that's odd as it seemed to work ok in detools.  What do I need to do to get a build to work on IE11 in browserstack?

edwh:
I'm using IE11 direct on my W10 home machine. What OS are you on?

JayyajGH:
Actually I've recreated it.    I only fixed it when it was on larger screens.  It looks like my fix has worked for what I intended but I also need to do something about the mobile view too.

I'm on Windows 10.  Oh damn. IE11 is available on W10 isn't it.  I forgot and I thought it was just Edge...

JayyajGH:
I'm just heading out so I'll take a look tomorrow. If you're happy with it on larger screens then this change is pretty standalone if you wanted to put it out which probably fixes the majority of IE11 use.  A very brief look at the mobile version might be a bit more refactoring work.

JayyajGH:
Just testing a fix for this now but about to head into a meeting...

JayyajGH:
Right.  See what you think of that...  I've basically just removed the auto margins and converted it to use flexbox.

layouts/default.vue
```diff
@@ -139,69 +137,69 @@
...
+              </div>
+            </b-nav-item>
+          </client-only>
+        </b-navbar-nav>
 
+        <b-navbar-nav class="">
+          <b-btn v-if="loggedIn" v-b-toggle.nav_collapse_mobile class="toggler white">
+            <v-icon name="bars" class="mb-1" scale="1.5" />
+          </b-btn>
+        <!--        <b-navbar-toggle v-if="loggedIn" target="nav_collapse_mobile" class="navbar-dark" />-->
```
aaronferris: Any point in this code if we're commenting it out? Looks like legacy but id be tempted to remove it, its just noise

JayyajGH: @edwh - Can this be removed or is it a reminder for something?

edwh: It can be removed.

JayyajGH: Done. :-)

edwh:
Looks good on Chrome, FF, Edge, IE11.  Will merge.  Thanks.

PR #217, opened 2020-03-12
Enhancements/profile image update

JayyajGH:
@edwh - See what you think of this.  I don't really like the solution as having a profile image at 93px and another at 100px seems crazy but it fixes it for now and at least it's easy to keep track of now it's using the profile image component.  I might look into sorting this header out next as it messes up at various sizes.  Then I could tweak this as part of that change.


edwh:
Better to use the component in a way you're less happy with than have the naked image instead, I reckon.  I'll merge.

PR #222, opened 2020-03-23
Enhancements/refactor notifications update

JayyajGH:
Move the notifications update process into the store to assist with the refactoring of the default layout.


edwh:
The basic move into the store looks fine.  I think it's ok to merge.  There is an issue which predates this, though.  On the one hand, we say "Get the list so that it will display zippily when they click".  But on the other hand when you click on the notification icon we clear the notifications and then fetch them, so you see the loader and it doesn't feel zippy.  I think I tried to do something about this before and then got distracted,because there's a larger refactor issue.  I'll spin that off into a separate issue.

edwh:
#223

JayyajGH:
Hopefully once some of this is broken up into smaller parts then it will make any fixes/refactors much easier.  I struggled for ages to work out how to refactor default.vue.  I found myself hitting brick walls every avenue I tried.  Hopefully taking notifications out (of which this change is part of that) into a component will start to make it easier.

edwh:
I write components that are too large.  A bad habit that is exacerbated when you're doing single-man projects.

JayyajGH:
Make it work and then make it right....  I'm sure given unlimited time and budget your components would have been perfect. :-)

edwh:
Ha.

PR #224, opened 2020-03-26
Refactor some of the notification logic into the store

JayyajGH:
Move some notifications logic out of the view and into the store with the getters.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -325,15 +325,9 @@ export default {
 
   computed: {
     notifications() {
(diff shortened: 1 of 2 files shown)


edwh:
Thanks. At some stage we really ought to shift to using mapGetters.

JayyajGH:
Yeah indeed.  Makes the code much simpler.

edwh:
I only didn't use them because my head was full at the start with getting used to Vue.  Mistake.

PR #231, opened 2020-04-04
Add a position to the chat badge parent so it sits in the right place

JayyajGH:


edwh:
What's the visual effect I should be looking out for her?

JayyajGH:
The number of unread chats should be displayed on the chat icon rather than over the freegle logo. See the slack thread for the issue.

PR #232, opened 2020-04-07
Feature/update password focus styling

JayyajGH:
Ok see what you think of this.  I've just done it on the settings page at the moment - hence the draft PR.  Because it's slightly kooky it would be good to maybe create a component for it rather than replicating it on the login page too.  I'll look into that next.


edwh:
Is the empty <label> doing something?

Visually it looks like what I want.  Probably this applies wherever we have an input add-on, doesn't it?  Like the email field above.

The earlier version of the similar fix was in Autocomplete.  You might or might not want to revisit that, if only to check that there isn't something in there which also needs to go in this.

I'm in favour of a component.  You might want to make sure you've seen SpinButton, which is a component for buttons with a busy indicator for actions; so perhaps we need a SpinInput which is similar.

JayyajGH:
"Is the empty doing something?"  - Did you mean the two empty end labels?  Nope... I'll remove them. Not sure how they got in there.

I'm still a bit unsure about the concept. A focus indicator is used to show which element currently has focus.  The idea of having a focus indicator around multiple, individually focused elements at once seems a bit odd.  I keep changing my mind of whether or not I like the concept.

Yeah it should probably apply to all similar fields.

Ooo SpinButton looks useful.  I was thinking that bit looked a bit messy in the element so that will tidy it up.

Looking at the autocomplete changes, I can't see anything in their that really does this.  Isn't it just a TODO?  Or maybe I'll need to look harder!

JayyajGH:
Also... where is "me.password" defined?  The same goes for properties like "me.onholidaytill".  Are they just being dynamically added (non-reactively) to the "me" object or am I missing something?

edwh:
me is added via a global mixin me.js.

I think with the focus, you're thinking about it from the HTML standards purity angle.  Just think about it purely visually.
* You have a thing which has a rounded oval boundary and is clearly a unit. 
* Then you click into and you get a shadow round part of it.
* The shadow is rounded at one side and not at the other.
* It seems to cut you off from the button that you thought was part of the thing you were clicking on.
* The show overlaps on top of the button making it look as though the input is somehow supposed to be vertically above the button, whereas you're used to shadows falling onto white space around something.

JayyajGH:
Yeah I think you're right.  My head is conflicted between the semantics and the visual.  It tests ok (to me) in a screenreader now and I think it's easy enough to use so I don't really have a problem with it.

So I'm guessing properties like "password" and "onholidaytill" are created dynamically when the API is auth/user is called?  Is that right?  I've not used Vue.set before.

edwh:
auth/user calls the API /session which returns a user object which goes into the store.  The current settings page is not a good example of how to do reactive stuff right, though, as I wrote it early on.  So I wouldn't look at that one too hard otherwise you'll make me blush.

JayyajGH:
I've just made a few tweaks and used the password component on the login modal as well.

I've made the PasswordEntry component the "owner" of the changed data.

For the settings page this means the previous password from me.password gets passed to PasswordEntry as a prop and then saved locally.  If the local value is updated then it is passed to the save API. This value isn't passed back to the parent.

For the login page this means the previous password gets passed to PasswordEntry as a prop and then saved locally. If the local value is updated then it raises an event to the login component for it to update it's local version using the .sync modifier.

Take a look and see what you think.  It makes the index.vue and loginmodal.vue files much simpler and keeps the complexity in the fairly small PasswordEntry component.  Am I missing anything?

I haven't properly tested this yet so it's still a draft PR

edwh:
Thanks - will have a look and get back to you.

edwh:
In PasswordEntry, it still has the saving/savedPassword fields that are now localised inside SpinButton.

But I'm not sure how this works in the two cases.  In Settings, we want you to be able to save a password via the input group append button.  But in login, that button shouldn't be there because it's the overall sign in/up button that triggers the action.  I've not tried it, so perhaps I'm missing something?

JayyajGH:
Ah right so that stuff can now be removed.  I squeezed the SpinButton stuff in and copied the implementation from elsewhere but didn't notice those fields were now replaced.  I'll sort that bit out.

What I've done at the moment is show/hide the save button with a prop so the login page doesn't display it.

edwh:
So you have. I am blind.  That;s ok then.

JayyajGH:
I have only briefly tested it at the moment though but I thought I'd let you give it the once over first to check you're happy with it before I tidied and tested it further.

edwh:
Yes.  As always, you are right to extract stuff into shared components.

JayyajGH:
Are you happy with the data passing methods?  I used this style when I first started using Vue before I discovered the store. In this case I think the store would be the wrong place for the data though and using props/events feels a better way.

edwh:
Yes.  In some places we use v-model, and I believe there is a very slight preference for using that where there is a single piece of info, but I don't feel strongly.

edwh:
The store would definitely be the wrong place for this.

JayyajGH:
I'm using v-model to keep the local data in PasswordEntry up to date with the input control but I mean for passing the data from child to parent. e.g. to the LoginModal.vue component.  Unless I've misunderstood your comment...

edwh:
So you mean use of .sync rather than emit()?

JayyajGH:
I'm still using emit to pass the data from child to parent.  But to get "fake" 2-way binding I'm using sync.  That provides the necessary abstraction so the child isn't actually directly updating the parent.  

The other option here I believe would be to do something like:
`:update:password="password = $event"`
which is a bit messier.

Not sure if I'm got the wrong end of your question though as I've seen you've used sync elsewhere...

edwh:
Sorry, my head is elsewhere this morning.  

Yes, what you've done looks fine. I'm not entirely comfortable with the way this kind of stuff works in Vue, but I'm not sure whether that's just my problem.  I believe it's changing a bit in Vue 3.  But the way you've done it looks ok.

JayyajGH:
Ah no worries, feel free to look at it at any point.  No rush.

I'll give it a tidy and a test and then give you a shout so you can take a proper look.

I find the React way even weirder as I think they use callback functions.  I guess all they are both doing is trying to define clear ownership of the data and creating control of who can update the single source of truth.  Which I suppose is why in a lot of cases the store is the right answer.  A single place to store the data and a single interface to update that data.

edwh:
Yes, except then the Store becomes a dumping ground, and you don't really know who updates it and when and whether or not in one component you will interact badly with another component.

There are changes in this area in Vue 3 (composition API) which I might get my head round when I have a chunk of time, like if society shuts down for a while or something.

JayyajGH:
Yeah good point.  Managing distributed data is hard!  I need to get my head around the Vue 3 stuff too.  Read a bit but still not looked into it deeply yet.

I think people are realising that they will never have time!

JayyajGH:
Something I've noticed when testing this is that on the settings page the saved password isn't brought through when entering the page so it is always blank.  I'll look at that separately to this issue as it is an existing problem.

JayyajGH:
@edwh - I think this is now ready to go. It's quite simple so not really much to test!  Probably worth you taking a proper look though as any issues in this area could obviously stop people logging in.  No rush though!

edwh:
I think that might be intentional, you can't retrieve the saved password 
from the server.

JayyajGH:
Yeah that seems sensible now I think about it! The browser can populate it if it wants but that will update the input field itself.

edwh:
I signed up, changed my password, logged out, logged back in.  So that 
seemed to work fine.

PR #234, opened 2020-04-22
Add underline to selected nav item

JayyajGH:
A small issue from the list...

Firstly I've added an underline to the selected nav item.  This fulfils the accessibility rule that colour alone shouldn't be used to signify something. If we struggle with a solution for the colours then this could be put live independently as it should help with usability.

I've tried switching the colours so that the selected item is white and the non-selected items are lighter but it's quite hard to see.  I've just done a contrast check on the regular colour and that doesn't pass AA either. i.e. #FFF on #61AE24 (white on green).  The green needs to be quite a bit darker for it to pass.  Do we have any room for tweaking the colours?  I'm guessing that green is the brand colour.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -201,51 +201,51 @@
         <b-navbar-nav class="ml-auto flex-row flex-wrap small">
           <b-nav-item class="text-center p-0" to="/communities" @mousedown="maybeReload('/communities')">
             <v-icon name="users" scale="2" /><br>
-            Communities
+            <span class="nav-item__text">Communities</span>
           </b-nav-item>
           <b-nav-item v-if="!simple" class="text-center p-0 white" to="/chitchat" @mousedown="maybeReload('/chitchat')">
             <v-icon name="coffee" scale="2" /><br>
-            ChitChat
+            <span class="nav-item__text">ChitChat</span>
           </b-nav-item>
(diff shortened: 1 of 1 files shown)


edwh:
Thanks.  Could you also do the mobile menu?

We could tweak the colours as long as it doesn't feel too different, 
i.e. green and fairly light.  What's the closest you can get that 
passes?

JayyajGH:
The mobile menu didn't appear to have any selection colouring previously so I left it as it was.  I'll take a look at that now though.
I think the green might need to be quite a bit darker before it passes.  I'll tweak it and send you a screenshot.

JayyajGH:
Actually the mobile view on desktop does have the selection colour but I'm not seeing it on my phone...

JayyajGH:
Mobile view now done too.

edwh:
I don't seem to see it either, so something odd is happening.

JayyajGH:
I've done it in the mobile view from devtools btw.  I can take a look at why it's not happening on an actual device.

edwh:
That would be nice.  Let me know if you want to release this PR and look 
at colours separately, or stick with this one.

JayyajGH:
To keep things simple it might be worth releasing this one independently as the colour issue and mobile issue might take a few discussions to resolve.  And this PR is an improvement I think.

edwh:
Agree - ready to go?

JayyajGH:
Yup go ahead

PR #235, opened 2020-04-22
Feature/fix nav button sizing

JayyajGH:
Sort out the help screen mobile layout issues as discussed in the slack thread.

### DIFF
--- a/components/Question.vue
+++ b/components/Question.vue
@@ -24,6 +24,7 @@ export default {
 
 <style scoped>
 .question__button {
+  /* Override the default button styling from global.scss */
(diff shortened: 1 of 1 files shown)


edwh:
Thanks.  There may be some other examples of this where we use 
b-collapse, but we can fix them if we see them.

JayyajGH:
Longer term we should probably get rid of the default button styling in global.scss but that might take a while.  Create a default button styling that is common and then have various variations of it.

edwh:
In retrospect given that most buttons have an icon inside it would have 
been better to start with our own button component.

PR #237, opened 2020-04-23
Fix the active nav selection colour

JayyajGH:
This should fix the issue where the active page wasn't highlighted on mobile devices.  The classes were a bit weird so I've just removed some and removed the manual use of the data- attribute.  I've tried this on Chrome and on a couple of mobile devices in browserstack and it looks fine.


edwh:
That shouldn't be in there, should it?

PR #248, opened 2020-04-29
Use grid to layout the header and remove the HTML duplication.

JayyajGH:
Re. the group header issue in the design shrapnel issue

Just been trying to remove the duplicated HTML in this component and due to the way the different header section are displayed in different positions in different breakpoints then flexbox can't be used to lay this out without duplicating HTML elements.  I can explain that further if you are interested.

This is where Grid comes in and is designed for this type of layout.  I usually end up using Grid for overall page layouts where you want to control where the sidebars, headers, footers are etc.

This PR is the refactoring of GroupHeader to use a grid layout.  The HTML is much simpler although the CSS is more complex due to the many different breakpoint layouts. There are 5 different layouts. They could possibly be simplified although for now I have replicated the previous layout.

The main issue here is Grid isn't properly implemented in IE11.  The result is the elements just get displayed as regular divs on separate lines.  It's usable but doesn't look the prettiest. I suppose the argument here is that people can't expect to use outdated technology and still expect the same experience. The counter argument is that people do. :-)  The other option here is to add in IE11 specific prefixes like "display: -ms-grid" etc.  That would get it looking the same.

Thoughts?


edwh:
I'm not up to speed on grid.  Guess I should change that.

I do care about IE11 - it's still in support and we have people using old kit (and so we should given our purpose).  So I like the sound of holding your nose and adding in IE11 CSS.

JayyajGH:
I've not used Grid a huge amount. The basics are probably even simpler than flexbox. Grid is much more regimented about the layout which is why it is great for overall page layouts.  I can give you a quick explanation of what I have used if we decide to go ahead with these changes.

The decision we made at work was we would make IE11 functional but not beautiful.  Everything would need to work but it wouldn't look as pretty as in other browsers.  I'm not suggesting that is necessarily the right solution for Freegle users but it's an option.  I'll attach a picture of how it looks in IE11.  If you don't like it then I'll add in the IE prefixes and see how it looks then.

The problem with IE11 was that MS implemented Grid early on but then the spec changed and they never updated it.  So it's still using the first version of the spec.

JayyajGH:
image

edwh:
That looks ok in IE11.  I can live with that.

JayyajGH:
That's the cool thing about browsers I guess. If they don't recognise something then they just ignore it so stuff just kinda works.

I'll check if anything needs tidying and do some more testing in other browsers/devices then.  I'll drop you an explanation of what I've done as well.

JayyajGH:
grid

JayyajGH:
So a quick explanation of what I've done...

I drew out the existing layout at all breakpoints and split it up into rows and columns. Grid will automatically guess the rows and columns if you don't specify them.  For the columns I have specified them using grid-template-columns.  I'm basically changing the number and width of columns are various breakpoints.

There are loads of different ways to lay stuff out.  I've used the method "start / end" like "grid-row: 1 / 3" which means start are grid line 1 and finish at grid line 3.

That's basically it.  I've used flexbox to handle the buttons within the grid as they move from row to column layout depending upon how much space they have.  I let flexbox work that out.

edwh:
Thank you.  I will need to read up on Grid a bit and then refer to this, 
I think, but I'm very happy to trust you on it if you'd like it 
released.

JayyajGH:
Grid and flexbox are definitely here to stay as the way to layout pages. Yeah if you wanna have a read then feel free to fire any questions at me.  The stuff I've done here is really just the basics but it can do so much more.

The main reason I used it here was to get rid of the HTML duplication.  The previous fix I made was required in two places; one for large screens and one for smaller screens with flexbox sorting out the medium areas.

I still need to do some more testing so I'll give you a shout when I have got around to it.  Should be in the next day or so.

JayyajGH:
Right so I've now tested on the following browsers/devices and it all looks OK.  IE11 and Safari older than v10 lay the sections out incorrectly in rows but it is still readable and useable.

I've tested on:

- Chrome on W10
- Edge on W10
- Firefox on W10
- IE11 on W7
- Safari 13 on Catalina
- Chrome on Catalina
- Safari 10.1 on Sierra
- Safari 9.1 on El Capitan
- Safari 13 on iPhone 11
- Safari 11 on iPhone 8
- Safari 10 on iPhone 7
- Safari 9 on iPhone 6S

Any other devices that would be worth checking?

edwh:
That's pretty comprehensive.  Perhaps an Android via Browserstack - do 
you have access?  I forget.

JayyajGH:
Oops sorry I did but it was on page 2 of my list.

- Chrome on Galaxy S20
- Chrome on Galaxy S8

These were all tested on browserstack.

JayyajGH:
Yeah I have the free Charity subscription for browserstack.  Thanks Browserstack. :-)

edwh:
Ok.  That's better testing than most fixes get, but probably justified 
since we're using a new CSS feature.  Thanks for being thorough.  Ready 
to merge?

JayyajGH:
It appears that Browerstack opens the last page you were on across devices so it's really quick to test the same page.  Yeah it's now good to go.  I'd probably be happier if you took a look on a couple of devices/browsers though.  I always hate testing my own work.

edwh:
I'll have a look.

edwh:
I can't make it look very ugly.  These are the worst, where the line break is in a slightly poor place w.r.t. icons or commas or there's more whitespace than it could have.

image

image

image

But it looks at least as good as the current one.  So I'm merging.

PR #249, opened 2020-05-07
If we are displaying a region then make sure it's the map focus

JayyajGH:
So I've just kept to the same method that you have used here for updating the map.  i.e. using the idle event.  "The internet" seem to set up a use-once event handler on idle instead but I think  I prefer what you've done here.

So basically I've just said that if you pass in a region then loop through the list of groups to get the bounds and then call fitBounds.  Pretty much the same as what you did previously.

### DIFF
--- a/components/ExploreMap.vue
+++ b/components/ExploreMap.vue
@@ -353,16 +353,38 @@ export default {
     idle() {
       const google = gmapApi()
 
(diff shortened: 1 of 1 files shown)


edwh:
We may end up pulling this out into a mixin for other map uses later - but we can do that at the time.  Will merge - thanks.

PR #250, opened 2020-05-11
Style the bootstrap checkbox

JayyajGH:
Here is one option with regards to making the checkbox more obvious.

I've just styled the bootstrap version to be a bit bigger.  This would need some messing with alignment of the text too but I just thought I'd show you to see if you think it's any better than what is there currently.

### DIFF
--- a/components/CovidPostWarning.vue
+++ b/components/CovidPostWarning.vue
@@ -1,9 +1,12 @@
 <template>
   <NoticeMessage variant="danger">
-    <p>
-      <b-form-checkbox v-model="currentCovid" size="lg" class="d-inline blah" @change="$emit('update:covid', $event)" />
-      &lt;--- Click here to say you consider this item essential and will follow the guidance on safe freegling during COVID-19.
-    </p>
+    <div class="mb-3">
+      <b-form-checkbox v-model="currentCovid" size="lg" class="d-inline" @change="$emit('update:covid', $event)" />
+      <span class="warning-text">
+        <v-icon name="arrow-left" />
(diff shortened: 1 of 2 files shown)


JayyajGH:
The other option would be to create our own checkbox style but that seems a bit overkill for a single checkbox and doing it the bootstrap way would keep the style consistent.

edwh:
I've not looked at this, but I don't think bigger is sufficient.  We 
understand what checkboxes are, but if you think about it from scratch, 
a blank empty square is not a compelling call to action for someone who 
(genuinely) might never have seen one before.

JayyajGH:
My background has always been creating stuff for tech-savvy users and so this is a new area for me!  

Personally I thought that what you had already created was really good.  It draws your eye and the message is clear what you have to do.  I can't think of any way of doing this any better.  The only thing in my head in this scenario is to attempt to go back to skeuomorphic basics and try and recreate a toggle switch of some kind.  Bootstrap provides one of the Apple style sliders which might be more meaningful.

image

edwh:
I wondered about that.  Decided against it.

Maybe just do a slightly nicer border, with shadow, check centring, make 
the arrow look less like ASCII art?

JayyajGH:
What about something like this?

This is actually really tricky!  The bootstrap checkbox is doing some weird stuff with positioning and font-sizes and so the usual alignment methods won't work.  This is a complete hack but it seems to work!

edwh:
That's probably neater.  I'll take it.

Yes, I was rather befuddled by what Bootstrap was doing.

PR #251, opened 2020-05-12
Better align the nav icons

JayyajGH:
This is an idea for how the navbar could be better laid out without crushing the longer text.  I've basically let flexbox deal with the layout.  This is currently untested but just to give an impression of how it will look.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -688,7 +688,7 @@ nav .navbar-nav li a.nuxt-link-active {
 
 .nuxt-link-active .nav-item__text {
   border-bottom: 1px solid white;
-  padding: 2px;
+  padding-top: 2px;
 }
 
 ::v-deep .nav-link {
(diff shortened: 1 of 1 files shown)


edwh:
You just think flex solves everything.

...because it does.  That seems much better.  I feel there must be a downside somewhere, but I've tried different sizes.

JayyajGH:
Ha!  It just means you don't have to worry about it yourself or set any explicit sizes.  If you add a new item or change the text etc. then flexbox will just deal with the alignment without you having to change anything.

I think spending the last 5 years working on a CMS where you didn't know how many items were in a list and things would just get added/removed by a content editor meant that I became a big fan of flexbox!

I'll tidy this up then and test it in various sizes/devices.

edwh:
You do have to care a bit about navbars though, because of the 
breakpoint where it switches to a dropdown menu.

JayyajGH:
Yeah I set a maxwidth on it so it still kept within the previous boundaries so it still worked with the switch.

JayyajGH:
Had a busy afternoon so I'll look at this tomorrow...

edwh:
I'll stop your pay cheque.

JayyajGH:
I've made a couple of tweaks and given it some testing and it looks good to me.

PR #255, opened 2020-05-17
External links

chriscant:
I'm adding an ExternalLink component and altering almost all of the external links to use it.
This should allow the iOS apps to open such links in the external Safari browser.
And fixed one wee bug along the way.

### DIFF
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -104,7 +104,9 @@
               Website
             </b-col>
             <b-col cols="8" md="9" class="forcebreak">
-              <a :href="event.contacturl" target="_blank" rel="noopener noreferrer">{{ event.contacturl }}</a>
+              <ExternalLink :href="event.contacturl">
+                {{ event.contacturl }}
+              </ExternalLink>
             </b-col>
           </b-row>
 
@@ -350,6 +352,7 @@ const OurFilePond = () => import('~/components/OurFilePond')
 const StartEndCollection = () => import('~/components/StartEndCollection')
 const NoticeMessage = () => import('~/components/NoticeMessage')
 const DonationButton = () => import('~/components/DonationButton')
+const ExternalLink = () => import('~/components/ExternalLink')
 
 function initialEvent() {
   return {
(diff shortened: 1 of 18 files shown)


edwh:
Looks basically fine.  Tedious to make all the changes - thank you.

Won't this have introduced whitespaces as part of the links in some cases?  They look messy when you move your mouse over to highlight them, so to avoid those I have put the <a>...</a> on a single line.  There are occasions when I need  to avoid the auto-formatting kicking in.  I try to avoid full stops in links too, which has snuck in for ModSettingsGroup and ModSupportAddGroup.

I'm very very slightly worried that the behaviour might break because of the event handler which wouldn't have been there before.  I'm not sure whether this is just cowardice, but we could probably avoid attaching the event handler at all by use of v-if and a computed property which said whether we were in the app.  

Does the change in ModMissingFacebook still preserve other btn behaviour such as mouseover colour change?  Probably.

chriscant:
Avoiding a blank before the full stops was why I put them inside. However I'll try to tighten up the formatting and avoid reformatting to make that work.
I'll make the event handler iffy.
I think the button behaviour still works, but will check.
I need to get the iOS app side working tomorrow, so I'll get that to work and do these changes.
I've previously had separate branch versions of any app-specific code but it would be good to start integrating properly.

edwh:
Ta.  It's a pain to get the integration done, but it increases the 
chances that I will start to make some app-related changes just as a 
natural thing, or at least be more aware of issues to warn you about.

chriscant:
I've made it so that external links don't introduce spaces.
I don't think you can have a v-if just on the event handler.
The alternative is to have two alternative A links.
There can only be one element in a component so I'd have to wrap the alternatives in a SPAN say, which is a bit grotty.
So I've not done that.

edwh:
Ta.  Will review, but I think it's possible to have conditional event 
handlers as per 
https://github.com/vuejs/vue/issues/7349#issuecomment-354937350

chriscant:
I'll work conditional that in then, later

chriscant:
If I change it to have a conditional click handler
`<a :href="href" target="_blank" rel="noopener noreferrer" v-on="{ click: isApp? openInBrowser : null }"><slot /></a>`
then Vue moans in the dev console:
`[Vue warn]: Invalid handler for event "click": got null`
and then does what we want.
But it doesn't feel like a very clean solution.
Just using a handler which returns true to bubble should be safe?

edwh:
Let's try it - enough paranoia pandering.

Ok to merge?

chriscant:
Go for it

PR #258, opened 2020-05-19
Tweaks to the layout on group header

JayyajGH:
Thought I'd add this in here for discussion.

I've moved the volunteer avatars and the button below.  Currently there are two options where one uses the bootstrap card and another with just a line break.

Thoughts?

### DIFF
--- a/components/GroupHeader.vue
+++ b/components/GroupHeader.vue
@@ -31,12 +31,14 @@
       </div>
       <div class="mt-2 group__buttons">
         <div class="button__items">
+          <!--
           <b-button class="mb-1" variant="white" :href="'mailto:' + modsemail">
             <v-icon name="question-circle" />&nbsp;Contact&nbsp;volunteers
           </b-button>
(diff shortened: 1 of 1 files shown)


edwh:
I think that is better and makes it look less cluttered on small screens.  

I prefer the version with an hr, rather than a card.  Add a colon after 'here'.

I still feel it needs a bit of something around the profile pictures, though.  You don't really know quite how many you're going to get, so there's something kind of incomplete about it.  I can't quite put my finger on it.  But this is better as it stands, so I'd take that.

JayyajGH:
I've removed the other card option to make it easier to talk about.  And added a colon...

JayyajGH:
I'll try and think of something else to help the profile pics.

JayyajGH:
I still need to tidy this as well as there is commented out code etc.  I'll tidy it now and see if I can think of any other options.

JayyajGH:
Ok I've now given it a bit of a tidy so I'll mark it as ready for review.  I can't currently think of any improvements to the pics but I can that could always be a phase 2 if we think of something.

PR #259, opened 2020-05-20
Feature/accessible buttons  no pointer

JayyajGH:
Button elements should not have a pointer.  These are for anchor elements.

If it's got a hand pointer then you should be able to right click on it to open in a new window.  That doesn't make sense with a JavaScript button.

The bootstrap default is to give both buttons and anchors hand pointers.  Bad bootstrap.

Adam Silver is a really good authority on forms and form controls.  This is one of a series of articles he wrote on the subject.  Technically we shouldn't be styling buttons and anchors and anchors as buttons but that has become fairly standard now so it's hard to get away from.

https://adamsilver.io/articles/buttons-shouldnt-have-a-hand-cursor/

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -138,6 +138,11 @@ button {
   white-space: nowrap;
 }
 
+.btn {
+  // Override the bootstrap default. The cursor on a button shouldn't be a pointer
+  cursor: auto !important;
+}
+
 .clickme {
   cursor: pointer;
 }
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -133,14 +133,12 @@ body.modtools {
   left: 0;
 }
 
-// Brought in from style.css
 button {
   white-space: nowrap;
-}
 
-.btn {
   // Override the bootstrap default. The cursor on a button shouldn't be a pointer
   cursor: auto !important;
+
 }
 
 .clickme {


edwh:
Hmmm.

"Like many people in the web community, I used to believe that the hand 
(or pointer) cursor signifies an element is clickable. "

...him, me and everyone else.  Because now, it does.

The argument seems to boil down to:
"To avoid confusion".  I don't buy that.  Removing this adds confusion.  
Without this, how do you tell if an element is clickable without 
clicking on it?  "maybe we need a new cursor" - but we ain't got one."It 
should look clickable anyway".  It should, but that is not an argument 
for not having the hand in addition.  If people aren't sure, they do 
actually move their mouse over it to see, where they are using a 
mouse."Buttons and links are different".  Only to geeks.  To users, they 
are both "a thing that does something if I click".  Whether a button 
takes you to a different page to do it, or does it in JS, is an 
implementation matter.
I have never had anyone complain that there is a hand on a non-link 
item.  I have had people complain that there isn't one on a non-link 
item.

This feels like prioritising standards purity above usability.

JayyajGH:
I'm heading out now but I'll read and respond later...

edwh:
It's a highly entertaining religious war, and I am grateful to you for 
drawing my attention to it.

Here's a stackexchange one:

https://ux.stackexchange.com/questions/105024/why-dont-button-html-elements-have-a-css-cursor-pointer-by-default

Of course, all the stackexchange actions have hand cursors.  But perhaps 
that's ok, because they're links really.  Ok, they're links with href=#, 
but users understand href=#, and so when you have href=# users don't 
expect to be able to right-click and open in a new tab even though the 
browser gives them that option :-).

Bootstrap took it out, then put it back, I think.  IMNSHO for good 
reason.

JayyajGH:
Yeah these things can definite cause a lot of debate!

I think the main reason I've always done this is because it's the default browser behaviour in (as far as i know) all browsers.  Create a <button> and the browser will give it a regular cursor, not a hand pointer.  Create an <a> and the browser will give it a hand pointer.  That is probably done as, just like in the old days, the anchor by default has weak affordance.

It was simple until people started making links look like buttons and buttons look like links!  I think the main advantage I can see is if it's got a hand pointer you know that you can then right click it to get a context menu with "open in new window" etc.  As to how many other people know that is another question. (I think I know the answer to that! :-) )

I can see it referenced with regards to WCAG guidelines but it doesn't appear to be listed explicitly.  So even though it's technically incorrect, it looks like it would probably still pass a AA audit so looks like the best method in this case probably would be to leave it as the bootstrap default.

edwh:
I'm happy with that.  If we want the aggro we can always go to a pub and 
say "What about those hand pointers on buttons, eh?"  Should start a 
good ruck.

PR #261, opened 2020-05-22
Rename the primary button to be secondary button.  Because it is.

JayyajGH:
The secondary button was called the primary button.  This is stage one of sorting this out.  I've renamed the primary colour and the primary button to be secondary.

I've looked at the main pages and this looks ok but I might have missed if it's been coded differently.

The next step will be to rename the success colour/button to be the primary.


JayyajGH:
A related question... should the send button in chat be primary rather than secondary?  i.e. green rather than blue?

JayyajGH:
I've now renamed the success button to the the primary button

edwh:
It should logically be primary. I was swayed by Facebook's colour scheme. Feel free to change it.

JayyajGH:
Done.  I've done the regular and the popup button

edwh:
Are you sufficiently sure that there isn't any lurking bootstrap behaviour which we haven't currently got in our own CSS which will then be lost?

JayyajGH:
The docs don't mention any specific behaviour from the button "variant".  It's only the "type" prop that is documented to affect the behaviour.  The only difference in the generated HTML is the addition of the styling class and that shouldn't be use to hook JavaScript off.
So I guess my answer is from the documentation it shouldn't affect behaviour and from clicking around on a few buttons they appear to work as normal.
Is that enough or would you like me to do some more investigation?

edwh:
I was wondering about hover classes or borders or that kind of thing.  
Maybe just have a look at the applied styles?

JayyajGH:
Bootstrap adds the "btn" class and then a class for the variant like "btn-success".  We override the "btn-variant" class in our css so we are in control of that.
Actually I've just noticed off the back of that that we aren't setting the focus styles so we are just inheriting the bootstrap ones.  I'll override them too so they are the correct colour.

edwh:
Ta.  There's a Love button on chitchat which turns a silly colour when 
you click it, which might be in that kind of area.

JayyajGH:
Which button is that?  The "Love this" button I am looking at is just the standard white one with a blue hover

edwh:
I notice it most on mobile.  It's white, and when clicked it goes blue.  
I then have a text-success spinner in it while the action happens, so 
you get green on blue which looks silly.

That should really be using SpinButton anyway - I've not changed 
everything to use that yet (search fa-spin).

But there's some mess around SpinButton and colours.  I introduced 
spinclass to help with that but it was a bad idea, and the colour of the 
spinner probably needs to be chosen based on the variant, and also 
change when the button colour changes (e.g. hover) rather than being 
fixed.

So SpinButton could use a bit of a look at to make sure the colours all 
work in the different states no matter what you do.  Then I could just 
go through and make everything use SpinButton.

JayyajGH:
I think that's just the regular hover colour of the "white" button.  On mobile it seems to retain it's hover colour whilst the spinner is working.  That should work as normal with my changes.  You can test it on desktop by continuing to hover over the love button after it is pressed and you get the same effect.

JayyajGH:
Is there an issue raised for the spinbutton issue you mentioned?

edwh:
Yes, I think it's the regular behaviour.  There isn't an issue for the 
SpinButton stuff.

JayyajGH:
The love button just looks the same as it did before in that case. i.e. still a bit weird with green on blue sometimes.  I'll raise an issue for the spinbutton stuff then and hopefully get around to looking at it!

edwh:
Yes, I agree this is unaffected.  Thanks.  Is this one ready to merge 
now?

JayyajGH:
Just noticed there are now some conflicts so I'll fix them first...

JayyajGH:
Conflicts now fixed...

edwh:
Ok, I'll grab it.

PR #262, opened 2020-05-24
Add component for email address

edwh:
There are various places where we ask for email addresses.  This adds a component which:

- Does syntax validation on the address.
- Makes use of a server API to suggest alternative domains to help with typos.
- Avoids placeholder, which we need to gradually retire for accessibility reasons.

It doesn't support the input group add-on usage that we have in several places, which means it changes things visually a bit.  I'm not sure whether it ought to.

It could do with another set of eyes to check it behaves right in the various places.


JayyajGH:
Still trying to look at this.  Every time I start I get distracted on something else....  I've got as far as checking it out though. :-)

edwh:
Checking it out, or checking it out?

JayyajGH:
The first one.  Or maybe the second one...

components/EmailValidator.vue
```diff
@@ -0,0 +1,147 @@
...
+  </div>
+</template>
+<script>
+import { required, email } from 'vuelidate/lib/validators'
+import { validationMixin } from 'vuelidate'
+import ValidatingForm from '../components/ValidatingForm'
+import ValidatingFormInput from '../components/ValidatingFormInput'
+import validationHelpers from '@/mixins/validationHelpers'
+
+const a = require('axios')
```
JayyajGH, line 41: What is the reason for the speific require of axios here?

components/EmailValidator.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <div class="d-inline-block w-100">
+    <validating-form
+      :class="{ 'justify-content-around': center, 'd-flex': center }"
```
JayyajGH, line 4: d-flex: center? Should that just be d-flex to set display:flex and justify to centre?

pages/find/whoami.vue
```diff
@@ -10,18 +10,11 @@
...
             <p>You will get emails from us, which you can control or turn off from Settings.</p>
-            <b-form-input
-              ref="email"
-              v-model="email"
-              type="email"
-              size="lg"
-              class="d-inline-block form-control email"
-              placeholder="What's your email address?"
-            />
+            <EmailValidator :email.sync="email" :valid.sync="emailValid" center />
```
JayyajGH, line 13: On  this page I'm getting "aria-describedby="undefined-feedback".  And the label associcated with it has a "for" of "email-"

JayyajGH: I don't think you need the aria attribute as long as your label for is correct

components/EmailValidator.vue
```diff
@@ -0,0 +1,147 @@
...
+    }
+  },
+  data: function() {
+    return {
+      suggestedDomains: [],
+      focused: false
+    }
+  },
+  watch: {
+    async email(newVal) {
```
JayyajGH, line 87: I'm not getting any suggesteddomains.  I haven't looked into why yet though.

components/EmailValidator.vue
edwh: None - will use this.$axios.

components/EmailValidator.vue
edwh: Yes.

pages/find/whoami.vue
edwh: Thanks.  This comes from ValidatingFormInput, which Nick wrote.  It looks like it implicitly (inside validationFieldHelpers) expects an id, but doesn't force that to happen in any way, and there are examples where it doesn't.  This probably needs some unravelling, if you fancy it.

components/EmailValidator.vue
edwh: You need to use a domain which is quite similar to a real one, e.g. btintrnet.com

pages/find/whoami.vue
JayyajGH: I'll create an issue for it now so it doesn't get forgotten then.

edwh:
Cheers.

components/EmailValidator.vue
JayyajGH: Yeah I'm still not getting any.  Might be me doing something stupid though.  I'm just gonna grab some lunch and then I'll take a deeper look.

edwh:

components/EmailValidator.vue
JayyajGH: I'm not getting the watch on the email prop running.  It is set once when the component is created but then when I type it's not run again.  Should it be running as I type on the give/whoami page for example?

edwh: You're right.  The watch works in the login modal, though.  Will dig some more.

edwh: I've pushed a fix without understanding the problem.  The case where it wasn't working was where we had a prop in the parent which was a computed property.  The set() method of that was firing, but the watch wasn't triggering.

I moved the logic out of a watch and it seems to work better.  But please test.

JayyajGH: I'll give it a go

JayyajGH: Now on the give page I'm not actually getting the "Freegle it" button until I change the existing email address.  Does that make sense?  I'll explain better if not...

JayyajGH: And then when I delete the last character of my email (jayyaj@hotmail.com) and it it back it again then I get the freegle button but it also has a suggestion "do you mean hotmail.com" which is what I've typed

edwh:
It makes sense, but I can't repro that.  Will look at the pointless 
suggestion.

components/EmailValidator.vue
JayyajGH: I get this on first entry to the page:
image

If I click into the email field and then out again then I get this:

image

If I click into the email field and change something and then I get this:
image

JayyajGH: Not sure if that helps

JayyajGH: Is it because "valid" is only being set "if (email !== this.email)" in the checkState function?

edwh:
I need to do a better fix where I understand why the prop isn't getting 
updated, I think, as there is a comment specifically saying that we 
can't do stuff in checkState because of vuelidate.


JayyajGH:
ok i'll ignore it for now then.  Give me a shout if you want me to re-test

edwh:
Better fix applied.

Yes, please re-test.  This is very valuable.

JayyajGH:
Superficially that looks much better.  I don't get the tick in the email field unless I actually enter it.  Is that correct?

edwh:
You mean if it was already present?  I hadn't specifically intended 
that, but it seems ok.  If your email address remains unchanged, which 
is the mainline case, you don't need the clutter.

JayyajGH:
Yeah if there was an email address already there.  I'd agree but I just thought I'd check.

edwh:
If you can think of any style improvements for the message, be my guest. 
  The warning colour is pretty unreadable, and the danger colour is too 
shouty.

edwh:
I'll merge this now so that I can put it on our test systems.

JayyajGH:
I think it looks quite nice visually but it isn't accessible.  But a lot of the other text on the page isn't either.
Fundamentally you need light colours on dark backgrounds or dark colours on light backgrounds.  A light colour like that needs a really dark colour background to be accessible.  I think I'm heading down the thought path that if you want the site to look light and fresh then you need a light colour background.  Therefore dark text is probably the only way to go.
Now the buttons are renamed I'll start to take a look at tweaking their colours to see if I can create something with a similar vibe that would fit it.  Then could then maybe drive tweaks to the rest of the site.

edwh:
I agree that a we need a light background for the vibe.

PR #268, opened 2020-05-27
First stab at more accessible button colours

JayyajGH:
This is a first go at some more accessible buttons.

I really don't like the look of the black text on the green or blue buttons and so this is keeping the white text and making the background colours darker.

Thoughts?

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -25,7 +25,8 @@
 // There are six types of button used on the site - primary, secondary, white, danger, warning and link
 // This is the primary CTA
 .btn-primary {
-  background-color: $colour-success;
+  //background-color: $colour-success;
+  background-color: #00823B;
   background-image: none;
   border-color: $color-green--dark;
   color: $color-white;
@@ -34,8 +35,9 @@
 }
 
 .btn-primary:hover {
-  color: $color-black;
-  background-color: $color-white;
+  //color: $color-black;
+  //background-color: $color-white;
+  background-color: #00703C;
   border-color: $color-blue--2;
 }
 
@@ -45,7 +47,8 @@
 
 // This is the secondary CTA
 .btn-secondary {
-  background-color: $colour-secondary;
+  //background-color: $colour-secondary;
+  background-color: #007E9E;
   border-color: $color-blue--dark;
   color: $color-white;
   text-shadow: 0 - 1 px 0 $color-black-opacity-59;
@@ -53,8 +56,8 @@
 }
 
 .btn-secondary:hover {
-  color: $color-black;
-  background-color: $color-white;
+  //background-color: $color-white;
+  background-color: #006E8A;
   border-color: $color-blue--2;
 }
 


JayyajGH:
I do realise that this goes against the comment I made on the other PR btw.  When I tried darker text on a lighter background it just looked weird.  Here I've tried to keep as much of the brightness/fun as possible whilst keeping the text white.

edwh:
They certainly feel clearer to read.

They also feel a bit sombre, but there may be no way round that, and I might just have to live with it.  I'm not sure this is very on trend, but is there scope for having the dark colour in the middle of the button to get the text contrast, but lightening the mood towards the edges using gradients?  

There's an argument that I've gone for vibrant colours in the buttons because I can't really do visual stuff.  So maybe we could keep the buttons sombre and clear, but have other design elements on the site which are more lively?

I think my main issue with these colours as they stand would be that they feel like "peers".  It's not visually clear that one is supposed to be the primary CTA and the other is secondary.

JayyajGH:
Yeah they do feel a bit more sombre.  And we would also need to do something about the colour of the  header as well which might accentuate that.

I'll have a play with a gradient.  The main problem I can think of there is on a smaller button there isn't much space around the text.  It might end up looking a bit like Windows 3.1 buttons.

This is where you need a skilled designer rather than a couple of developers having a go! :-)

JayyajGH:
Another option would be something like this:

image

JayyajGH:
Or this:

image

edwh:
:-).

We could have additional classes for the size (as well as the standard 
size= thing), perhaps.

JayyajGH:
Or a combination of it all.  The dark sombre green for the primary CTA, a paler button like in my last example for the secondary and then the existing white button for the tertiary.

edwh:
I'm not sure.  I don't know whether it looks button-y enough to make 
people click.  I don't know why I think that, though, so I could be 
wrong.

JayyajGH:
Yeah changing the button and text size in the button is another option. Although that could get tricky on mobile as we have a lot of buttons on some screens that would all need to be small to fit.

JayyajGH:
The general public are now pretty familiar with the new material design style flat design.  That doesn't necessarily mean that the freegle audience is though...

edwh:
I think it's best to assume a five year lag.

JayyajGH:
Would it be worth a bandit test maybe?  Or don't you have a focus group of volunteers that you use?  Or did I make that up?

JayyajGH:
There is an example of a combination of both colour styles for primary/secondary/tertiary.  Not sure if the blue blends a bit too much with the headers.

image

edwh:
It would indeed be worth a bandit test.  I'm not sure whether that 
should be on a single page or cross site.  Single page would be easier, 
though inconsistent.

edwh:
I think it's too subtle.  One trick I use to simulate cognitive or 
visual obstacles is to look slightly to one side, defocus my eyes, and 
count how many things I can see which are obviously clickable.  Or go 
cross-eyed.  I've no idea whether that's a legit technique, but in this 
example it would be just the primary.

For at least the primary and secondary I think we need a significant 
block of colour.  I think the secondary you have below would actually 
work best as the tertiary - and better than the current one - but with a 
slightly softer border.

JayyajGH:
I use the squint test which is pretty much the same thing. i.e. just squint your eyes.  I've also seen people add a blur to the whole webpage for a similar effect.

So are those colour examples too sombre? I can try and tweak them to make them as bright as they can possibly be but I think they will be similar.  Do you just want to have a think on it for a bit?

edwh:
I'd like a clearer distinction between the primary and secondary, if you 
can manage that, so the primary should be brighter.

I can imagine having tasteful rainbow borders.  And perhaps pink 
unicorns.

I'll do a bit of googling to see if I can find something I like for 
inspiration.

JayyajGH:
Remember when you're looking, that the majority of the internet isn't accessible.  Even the big sites like Amazon have shocking issues.  Once the dominoes case completes then that is likely to change fairly rapidly!

JayyajGH:
image

edwh:
I'm really not hating that, you know.  Though maybe curved, and it might 
be better to aim for that one to become primary.

JayyajGH:
Not entirely sure if you are joking or not...

edwh:
I've been thinking about other sites that our demographic might use.

It would be great if there were analytics to tell you that.  I can't 
find anything like that on Google though.

Mumsnet would be one.  I like their purple, though I'm not sure that 
would work with green.

Facebook, obviously.

Nextdoor has a green scheme, with a bright green primary button colour.

Money Savings Expert has quite a sombre primary blue colour, also uses a 
purple.

Greenpeace has a bright green button colour.

So there do seem to be some significant sites with bright green buttons.




JayyajGH:
The Nextdoor and greenpeace ones aren't compliant.

The mumsnet purple is really compliant.  Even passes AAA.  Their blue passes AA.  The MSE site passes AA too for both blue and purple.

I'll try and find the brightest colour we can get away with and see how it looks.

edwh:
I'm not joking.

You have to remember the demographic.  Obviously here we're into 
cliches, but lots of emojis, predominantly female, bit hippyish.

JayyajGH:
This is the brightest looking dark green I can come up with.  It's used elsewhere on the site too.

image

JayyajGH:
Another option for maybe creating a better hierarchy would be lozenge-ifying the buttons:

image

JayyajGH:
So with them all combined:

image

JayyajGH:
I'm not going to be around much until Monday now...  A bit tomorrow maybe...

edwh:
There seems to be some pixelation round the letters - is that real?

I still struggle with it.  Anything that dark doesn't work for me as a 
primary CTA.  The secondary colour is more eye-catching.

Let's turn it round.  Forget the existing colour-scheme for a bit.  Pick 
a colour that is a) bright and engaging and b) accessible.  What would 
it be?

edwh:
I think we could replace the white buttons with the colour and border 
approach you have there.  Check if it looks ok on a few other pages, 
like chitchat.  Then maybe commit that one so that it's out of your head 
and we're just looking at primary and secondary.

Not mad on the lozenges.

JayyajGH:
If you compare them side by side then they look about the same re. pixelation.  It does seem to be quite a pixelated font on Chrome.

JayyajGH:
Fundamentally with white text the colour needs to be quite dark.    Maybe we switch it around and try some lighter, brighter colours with dark text.

Here is the purple colour stolen from Mumsnet:
image

edwh:
That's weird.  I can zoom right in and see no pixelation:


JayyajGH:
image

edwh:
Fortunately if we did steal that it's not like there'd be a public audit 
trail showing it.

Yes, let's try some dark text.  I don't like black, but there might be 
other dark colours which would work.  The sombre feel might not come 
across if the colour is used for the text.

JayyajGH:
Yeah I'm getting less pixelation when i zoom in too

JayyajGH:
OK I'll try that next.  Dark text with a lighter green/blue.  I've just been using #000 for the examples but a gray might look better.

edwh:
Well, let's ignore the pixelation for now.

JayyajGH:
image

JayyajGH:
#77D42B background and #454545 foreground if you want to have a see how that feels in your own browser

JayyajGH:
Here is another stab at the primary/secondary colours.

image

I think we once again have the issue of the secondary standing out more than the primary.  But maybe they did before and we have just got used to it?  Having the light blue background for the tertiary button ends up looking a bit weird as the header bars are pretty much the same colours.

I'm wondering if we are just going around in circles here. I'm not really happy with anything that I am coming up with.  Is there the option for a shout out to designers to try and get some assistance?  I might try giving a girl I used to work with a shout again to see if she can help.  I asked her before but she didn't have time but she might have an hour to spare.  She used to work for an accessibility company too!

edwh:
Gosh, this is hard work. I admire your persistence.

I don't think we're coming up with much that's better.  I'm not sure 
we've fully explored the idea of light colours, borders, and dark text.

But it feels as though we might settle on something through exhaustion 
rather than because we really like it.

So yes, try your contact again on the off chance.  If nothing else, 
it'll give you an excuse not to have your head stuck in a colour chart 
for a while...

JayyajGH:
Yeah it is hard work!  It's easy to come up with something that's accessible but to also make that fit in with the vibe of the site is another matter.
My contact says she will take a quick look - but that doesn't necessarily mean that you will like the results!  Let's see...  Otherwise maybe let's park this for a bit and come back to it later.

JayyajGH:
Just noticed an oddity on the live site with regards to this conversation.  Right click on any of the anchors styled to look like buttons and the colour goes odd.  We need to set the active state colours too.  I'll add this to another PR to go with the primary/secondary renames.

edwh:
Ok, ta.  I've fixed one with the notification button on mobile.

JayyajGH:
Ah I saw that one and looked at the classes added by bootstrap but the btn- classes weren't there but I didn't look at mobile.

JayyajGH:
image

JayyajGH:
Another stab at something different.  I was wondering if the buttons should be a different colour from the main nav as it can distract from them.  All those button colours pass AA

edwh:
I'm not sure what I think about these, so I let it sit for a while.

They're not bad.  Probably the two colours are a bit close together.

I can see your argument about the nav in theory.  I have quite a strong 
attachment to the Freegle = green = good = do this kind of feel for the 
primary button, though.

The primary button on Facebook is also their brand colour:


Twitter's primary button is their logo colour:


Unless I've happened to stumble on outliers, this suggests your primary 
colour should be your brand colour?

JayyajGH:
I've never thought about the brand thing before but that's very interesting.  And makes sense.
On a plus side I think that is a decision about what colour the primary CTA should be. :-)

I spoke to my friend last week and she's mentioned a few improvements but got a bit distracted with other stuff and didn't really conclude anything about the buttons.  Her only comment around them was if you want to keep the colour then go for a ghost button style with the green border.  After I mentioned the customer base she then said that a look at what similar targeted sites do (much like what we did) - she mentioned tesco, hobbycraft and the daily mail.  She missed out mumsnet.  :-) 

Looking at Tesco they are using part of their logo/brand colour for their primary CTA but it happens to be dark and passes AA and AAA.

So conclusions so far...
1. The main CTA should be the brand green colour
2. The green isn't accessible with white text but we don't like it with black text.

Thoughts on achieving this:
1. Change the brand colour to a darker green (with white text) or a lighter green (with dark text)
2. Go for a ghost button style

Can you think of any other options?

edwh:
We can tweak the navbar and primary colour a bit as long as it is a very 
vivid green.  From what we've tried so far I would be surprised if you 
can get a white that contrasts enough with that.

I worry that a light button with a border fundamentally won't look 
buttony enough for the audience, unless it has rainbows and unicorns.

I would be ok with a button that had a dark green in the text section 
(to contrast with the white text) but gradients to the vibrant brand 
colour at the edge.  Kind of like the logo does, only the other way 
round - the logo has a brand colour in the middle but transitions to a 
dark green at the edges.

That doesn't match current preferences for flat designs, so it might 
look a bit crass to design folks - but probably not for our members - 
and might keep a brighter feel.

I don't know how to implement that myself, and it might be that we need 
to have classes to hardcode the sizes of buttons so that we could do it. 
  But let's ignore that for now and see if we can get one that works on a 
couple of buttons.

Another option might be to link this to the simple/full website toggle.  
But the trouble with that is that the first impression people will have 
is of the simple version, and therefore we lose the brand.

JayyajGH:
I'm going to park this in my head for a few days and then come back to it I think.

edwh:
Can't blame you.

Next project I do, I'll choose a brand colour bearing accessibility in mind!

JayyajGH:
You'd be amazed (ok maybe you wouldn't...) at how many design agencies don't bear it in mind!

Read a nice article today regarding how accessibility isn't like drowning - bear with me on this one...
Drowning is binary.  Whether you are at the bottom of the river or 1cm from the top.  You're still drowning.  With accessibility, any small improvements you can make will allow a wider audience to use your website.  Just keep chopping away at the small stuff.  They mentioned colour contrast first and then keyboard navigation second.

JayyajGH:
@edwh - Shall I close this PR as we didn't really conclude anything.

edwh:
Yes.  I think the new button work will subsume this.  Ooh, I used the 
word 'subsume'.

PR #273, opened 2020-06-03
Feature/accessible email field

JayyajGH:
As discussed in the associated issue.

Fundamentally this is to resolve the accessibility issue with the label "for" not matching the input element.

### DIFF
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -264,7 +264,6 @@
               />
             </b-form-group>
             <EmailValidator
-              id="contactemail"
               ref="email"
               size="md"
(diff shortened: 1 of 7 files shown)


edwh:
Thanks.  I'm delaying reviewing this as I want to get some other stuff live first.

JayyajGH:
There is actually an issue with the email field validation here.  When trying to add a volunteer opportunity if you don't enter an email address and click "add" you get no feedback but it won't add.  This is happening in live so nothing to do with this change so I'll look at it separately.

edwh:
Thank you.  The same might apply for community events, since the code is 
very similar for both.

PR #275, opened 2020-06-05
Accessibility and HTML improvements

JayyajGH:
Various accessibility and HTML improvements to the volunteer sidebar

- Remove block element within an inline element (h4 within a)
- The tabbing order was incorrect due to the button being floated
- The focus rectangle for the header was around the header and the button
- Add a meaningful aria label to the button rather than "Add"

### DIFF
--- a/components/CommunityEventSidebar.vue
+++ b/components/CommunityEventSidebar.vue
@@ -2,25 +2,29 @@
   <div>
     <b-card variant="white" no-body>
       <b-card-body class="p-0">
-        <b-btn variant="white" class="float-right m-1" @click="showEventModal">
-          <v-icon name="plus" /> Add
-        </b-btn>
-        <nuxt-link to="/communityevents">
-          <h4 class="pl-1 pt-1">
-            <v-icon name="calendar-alt" scale="2" /> Community Events
-          </h4>
-        </nuxt-link>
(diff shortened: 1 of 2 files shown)


edwh:
Have you checked CommunityEventSidebar for similar stuff?

JayyajGH:
oh yes, looks like it has the same issues.

JayyajGH:
Is there an easy way to add community events without hacking them in in javascript?  I don't have the sidebar shown because there aren't any existing events.  Is that correct?

edwh:
There's a fix I've done recently to force that sidebar to show, I think. 
  But you can go to the Events page from the navbar, and add from there.

JayyajGH:
I've now made the same changes to the other sidebar now too. 

There is still work to be done in this area.  The focus indicator isn't very good so needs a new colour and shape.  Also the events aren't very accessible in the sidebar.  Missing tab stops, non-accessible labels etc.  We should probably look into refactoring the common code between these out too.  I might look at some of those things next.

This is one good to go though.

edwh:
Ta.  I'll review today.

edwh:
When you do look at these, be aware that there is at least one difference, which is that events must be on a specific group, but volunteer ops need not be and can be system wide.

PR #276, opened 2020-06-08
Refactor the common parts from the sidebar items

JayyajGH:
So I've pulled out the duplicate code in these two components.  I don't think the thing you mentioned on the other PR is relevant here as the separate modals handle that.
The only thing I'm not happy with here yet is the name of the component and the names of some of the props.  Basically I need a name for something that can be an event or an opportunity.  Any ideas?

### DIFF
--- a/components/CommunityEvent.vue
+++ b/components/CommunityEvent.vue
@@ -6,9 +6,7 @@
           <span v-if="!summary" class="float-right small text-muted">
             #{{ item.id }}
           </span>
-          <span v-else>
-            {{ item.title }}
-          </span>
+          {{ item.title }}
         </nuxt-link>
       </b-card-title>
       <b-card-body class="p-1 pt-0">
--- a/components/CommunityEvent.vue
+++ b/components/CommunityEvent.vue
@@ -2,12 +2,14 @@
   <div>
     <b-card variant="success" no-body>
       <b-card-title class="bg-info pl-2 mb-0 pt-2 pb-2 text-truncate">
-        <nuxt-link :to="'/communityevent/' + event.id">
+        <nuxt-link :to="'/communityevent/' + event.id" class="event__link">
           <span v-if="!summary" class="float-right small text-muted">
             #{{ event.id }}
           </span>
+          <span v-else>
+            {{ event.title }}
+          </span>
         </nuxt-link>
-        {{ event.title }}
       </b-card-title>
       <b-card-body class="p-1 pt-0">
         <div v-if="summary">
@@ -145,10 +147,16 @@ export default {
   }
 }
 </script>
-<style scoped>
+<style scoped lang="scss">
+@import 'color-vars';
+
 .square {
   object-fit: cover;
   width: 200px;
   height: 200px;
+}
+
+.event__link {
+  color: $color-blue--2;
 }
 </style>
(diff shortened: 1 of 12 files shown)


edwh:
"CommunityFeature"?

JayyajGH:
It's not giving me one of those "that's it" moment but it's better than what I had.  We can always change it later if we think of something better.

edwh:
It's that or AbstractActivityOrLocationBasedThingummy.

JayyajGH:
Perfect.  You've hit the nail on the head with that one. :-)

JayyajGH:
We regularly have the conversation at work that naming stuff is the hardest thing about development.

edwh:
Yeah.  That and my thumb.

JayyajGH:
Is the key used when displaying the volunteer ops/community events important?  At the moment one uses "key-<id>" and the other uses "volunteering-<id>" .  Can it be changed to something like "item-<id>"?

edwh:
I think the key needs to be unique on the page otherwise Vue complains.  
So they need to be different otherwise if they are both displayed on 
sidebars it might clash if the ids happen to be the same.

"key-" is a terrible choice of key, though.  As long as it is unique 
across both types, "item-id-type" or something would be fine.

JayyajGH:
One doesn't use key - my mistake.  It is "event-" and "volunteerings-"

JayyajGH:
OK so the ID's ca n potentially be the same.  In that case I'll make them different

JayyajGH:
OK I've made some tweaks to naming now.  It's not the prettiest of components but it has at least removed the code duplication for the two sidebar items.

JayyajGH:
I don't have any community events.  Do you have any that you can check this bit works too?

JayyajGH:
Comment moved from another issue:

Sorry, I know you asked me explicitly about this, but reviewing the code I think the title should probably be clickable too, as I can definitely imagine people doing that to find out more.

JayyajGH:
Comment moved from other issue:

I underuse :is, I think, so it's nice to see that. It's a little bit undermined by the if statement with explicit reference to the value of featureComponent. Should we pass in a prop which we used to set the key value?

If it would help we could change VolunteerOpportunity and CommunityEvent to take the same prop name, e.g. item.

JayyajGH:
The title is now clickable

JayyajGH:
With reference to your comment around "is".  This is the bit I was referring to when I said it wasn't pretty.  I didn't want to change the prop on each component hence why I used the if statement.  But yeah I agree it would be nicer.
I'll change the components to use the same prop and pass in the key.

edwh:
I think you need to check for other uses of CommunityEvent and VolunteerOpportunity, and change to use the item prop in there?

JayyajGH:
Argh.  I think working on two changes in exactly the same area has confused me.  I've also got conflicts with my other change now so I'll sort.

JayyajGH:
That issue should now be resolved.  I've tested the regular places the VolunteerOp component is used.  Would you be able to test the communityevent component and the mod tools?

edwh:
Yes, I'll test those two now.

edwh:
Worth testing as I found a bug in a recent fix I made.  Merged.  Thanks for tidying this up significantly.

PR #277, opened 2020-06-09
Feature/anchor changes

JayyajGH:
Accessibility tweaks to the two sidebar items.  Only the more info button should be clickable.  Give the more info button better text for screenreaders.

### DIFF
--- a/components/CommunityEvent.vue
+++ b/components/CommunityEvent.vue
@@ -42,7 +42,7 @@
             </div>
           </div>
           <div class="text-center mt-2 mb-2">
-            <b-btn variant="white" size="sm" @click="showEventModal">
+            <b-btn variant="white" size="sm" :aria-label="'More info about ' + event.title + ' event'" @click="showEventModal">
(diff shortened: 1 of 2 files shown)


edwh:
Sorry, I know you asked me explicitly about this, but reviewing the code I think the title should probably be clickable too, as I can definitely imagine people doing that to find out more.

edwh:
I underuse :is, I think, so it's nice to see that.  It's a little bit undermined by the if statement with explicit reference to the value of featureComponent.  Should we pass in a prop which we used to set the key value?

If it would help we could change VolunteerOpportunity and CommunityEvent to take the same prop name, e.g. item.

JayyajGH:
OK I'll make the title clickable too.  Actually what might be easier is doing that as part of the other change where i've moved it into a component so I don't have to do it twice.  I'll add your comment on to that PR.

JayyajGH:
I think your other comment applies to the other PR rather than this one too so I'll move it across

PR #278, opened 2020-06-11
Make it obvious if the email dropdown has a positive selection

JayyajGH:
What do you think about this?

The slight delay in the Vue model (Vue view model????) getting updating actually provides a slight transition so it actually feels quite smooth.  The only slightly weird thing is the blue focus indicator doesn't go well with the green colour but that's a completely different issue!

### DIFF
--- a/components/SettingsGroup.vue
+++ b/components/SettingsGroup.vue
@@ -5,7 +5,7 @@
         <b-form-group
           label="OFFER and WANTED posts:"
         >
-          <b-form-select :value="emailfrequency" @change="(newval) => outcast('emailfrequency', newval)">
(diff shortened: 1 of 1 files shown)


edwh:
That seems fine.  You're right, there is a delay anyway.

We're not very consistent about how to set variable classes.  Sometimes we do it with computed props, sometimes with ?...:, and sometimes with the object version where you have the classname and a boolean e.g.:

`          <b-card border-variant="success" :class="{ 'ml-auto': !amUser } ">
`
...but I don't much mind, and I've used all of those.  I'll merge this - thanks.

JayyajGH:
I tend to switch it about too.  If there is already a lot going on on the element or it's slightly complex then I use a computed prop.  With further thinking this might be better as a computed prop as it explains what it's doing better.  I might switch it...

PR #280, opened 2020-06-12
Feature/make showlove tabable

JayyajGH:
The showloves button wasn't accessible by keyboard.  It's didn't have a tab stop or any focus indicator.

I've just turned it into a real button.  I'm still not sure if the accessibility side is correct yet.  The button is effectively a label showing a count and also a button to display some more info.  I think it's better than it was before but might need some tweaking in the future.

### DIFF
--- a/components/NewsLoveComment.vue
+++ b/components/NewsLoveComment.vue
@@ -16,10 +16,10 @@
           <v-icon name="comment" /><span class="d-none d-sm-inline">&nbsp;Reply</span>
         </b-btn>
       </li>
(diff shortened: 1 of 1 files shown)


edwh:
That looks better, plus we get the mouseover colour.

There's a similar issue in NewsReply, though the loves are not a separate component there.

JayyajGH:
I'll take a look at the NewsReply one too then.  I've basically just made it a button then you get all the button and tabbing stuff without having to do anything.

PR #282, opened 2020-06-16
Feature/accessible chat button

JayyajGH:
Turn the "naked" message button into an actual button so it can be accessed by keyboard and screen readers etc.

I've removed the high-jacking of the size prop for "naked" and passed in separate prop to use to specify no icon. That way we can use the standard bootstrap variant and size props to set up different styles. The naked version is only used in one place so I've defaulted the component to have an icon so I don't need to change it in loads of places.

There are still a few niggles with this section like the bullet is in some of the hovers but not others. I'll see if I can make some more improvements without too much work.  These options would be more accessible if they were in a "ul" for example but that might involve too much refactoring for now.

### DIFF
--- a/components/ChatButton.vue
+++ b/components/ChatButton.vue
@@ -1,12 +1,12 @@
 <template>
   <span>
-    <span v-if="size === 'naked'">
-      <span class="d-none d-sm-inline-block" @click="gotoChat(true)">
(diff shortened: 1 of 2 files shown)


edwh:
That seems neater.  Thank you.

PR #284, opened 2020-06-17
Fixes to make the light button work in other cases

JayyajGH:
OK so I've now created a permanent button variant for this.  I've given it a test in the areas you found problems in but I will continue some more testing this afternoon.

There is one niggle I'm found which is when you hover when the button is on the blue background it is still gray.  Trying to have a see-through button which has a context sensitive hover colour would be awkward.  The only thing I could think of would be to use some kind of filter on the background colour.  So I guess it depends on how much you hate it!


edwh:
The hover is fine.  Thanks.

PR #285, opened 2020-06-18
Add accessible label to showlove button

JayyajGH:


edwh:
This reminds me of something else that flickers in and out of my mind from time to time.  I've seen things in Bootstrap-Vue about how you should add an id for accessibility.  For example:

`To enable auto-generation of aria-* attributes, you should supply a unique id prop to <b-form-group>. This will associate the help text and feedback text to the <b-form-group> and, indirectly to its input control(s).`

We don't do this on a lot of of our form groups, though.  And you've added an aria-label here, but not on other buttons.  

I'm left with a feeling that we (or boostrap-vue) are lacking eslint rules, or something, to enforce good practice.  So I will continue to wreak havoc while you follow me with your screenreader picking up the pieces.  

Is there anything we could do about this?

JayyajGH:
This was definitely done in the work that Nick did as I made some fixes to it recently.  There were duplicates being created.  I think it was with the email field that definitely has it.  The input field needs to be hooked up to the label (usually with a "for" attribute) and the error field needs to be linked up to the input field (usually with an aria-describedby).  

Although I'm not expert, I would say that most of the time you don't need aria attributes.  Aria attributes are usually used when you are doing something weird and non-standard.  For example a regular <button onclick="callme()">Click here</button> wouldn't need them as a screenreader would read out "Click here button" anyway.  I've added them here as the only text on the button is the number of likes so it would just read out "2 button" without it.

This stuff is hard to lint as in most cases the linter won't know.  How would it know that "2 button" doesn't mean anything and instead it should be "2 likes" or something similar.  Although linters could pick up stuff like a missing "for" on a form label.

When I write new stuff I usually just check that I can use it correctly using the keyboard and then the a screenreader reads out usable information.  Finding issues is actually really easy and doesn't really need any knowledge of accessibility. Fixing it can be harder.

edwh:
Ok, let's ignore the aria stuff.  But in the example of components which 
BV says you need an id for accessibility?  Perhaps we just ignore it cos 
it's their fault for not making them mandatory.

I should stop procrastinating and do some keyboard/screenreader stuff.

JayyajGH:
I think they say you need an id so they can hook up the error.  In ValidatingFormInput it looks like we are doing that manually though.  I'm not sure why.  Maybe because we're using Vuelidate?

I'd definitely recommend trying to use the site using keyboard only and then with a screenreader.  Just spend 30mins having a play and you'll soon here how hard some bits are to use.

edwh:
I'll see if any volunteers want to do that.  Partly out of laziness, 
partly because it's something they could definitely help with.

JayyajGH:
I think it helps people write more semantic HTML though by just being aware of it so I think it's good for developers to know the basics.  Knowing that by adding a click event to a "span" means that you will then need to do a whole bunch of aria work just to make it work means that you're more likely to make it a button because it's easier and quicker and will likely just need a bit of styling.

Accessibility isn't hard.  It comes for free if you write proper semantic HTML.

Accessibility rant over...

edwh:
That click handler comment is gold.

...does regex search in the code for span.*click

Oh dear.

JayyajGH:
:-)

Every year there is a CSS free day when a lot of cool (usually tech) companies remove their CSS for the day.  That really exposes any weird stuff that's going on.  Suddenly your buttons won't look so much like buttons.

I'm guessing you've done it in a lot of places because you didn't want all the bootstrap stuff that comes with a button.   You've seen the hassle that I've had to make it work as an actual button.  A span just works as you want it.

PR #286, opened 2020-06-22
Remove the double link/button

JayyajGH:
This was previously a button contained within an anchor so it had two tab stops.  I've now made it the same as the one on the give page


edwh:
Ah yes.  Initially I didn't realise you could have "to" on a button, so there might be others of those.

JayyajGH:
The only problem it seems to cause is when you are tabbing through the page you get a double stop on this page. I've guessing it would confuse a screenreader too. These should probably be submit buttons instead of anchors but that's another story...

PR #287, opened 2020-06-23
Sort out the alignment of the email fields

JayyajGH:
See if this works for you.  I've changed the email validator to be in a column so it lays out correctly.  Most uses of it (like on the settings page) will lay out left aligned.  On the give and find pages I've aligned them centrally in the parent.


edwh:
That looks good.  Ta.

PR #288, opened 2020-06-23
Tidy up the tab stops

JayyajGH:
Having the title and the id connected just seemed very weird with the distance between them.  Hovering over the title also highlighted the id which was quite confusing.  I don't think this really works so I've split them up into two.

I've tested the volunteer op when it is and isn't a summary.  I've also updated the event but I don't have any to test so would be useful to have a quick check that it works ok.

### DIFF
--- a/components/CommunityEvent.vue
+++ b/components/CommunityEvent.vue
@@ -1,12 +1,12 @@
 <template>
   <div>
     <b-card variant="success" no-body>
-      <b-card-title class="bg-info pl-2 mb-0 pt-2 pb-2 text-truncate">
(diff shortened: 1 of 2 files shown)


edwh:
I added a test event on FreeglePlayground (you can do that).  Looks ok.  Thanks.

JayyajGH:
I can't see your test event on FreeglePlayground.  I've also tried adding my own to FreeglePlayground but it says it needs approval.  Can you approve that?

edwh:
Done

JayyajGH:
Perfect.  I can see that.  Thanks!  Did you delete yours?

edwh:
I think it just aged out.

PR #290, opened 2020-06-29
Resolve the scoping issues of the postcode styles

JayyajGH:
This should resolve the issue raised in slack re. the width of the postcode autocomplete.

There appears to be a difference in the scoping data- attributes when directly navigating to the give page.  I have now moved the class which is used to correct this position from the postcode component into the autocomplete component.  I think this is actually a better place for it to live as it is closer to where it is actually used and makes more sense as it overrides a property in that component anyway,

### DIFF
--- a/components/Autocomplete.vue
+++ b/components/Autocomplete.vue
@@ -635,6 +635,10 @@ export default {
   right: 48%;
 }
 
+::v-deep .postcodelist.autocomplete ul {
+  position: initial;
+}
+
 /*.autocomplete ul:before{*/
 /*content: "";*/
 /*display: block;*/
--- a/components/Postcode.vue
+++ b/components/Postcode.vue
@@ -206,8 +206,3 @@ export default {
   }
 }
 </script>
-<style scoped>
-::v-deep .autocomplete ul {
-  position: initial;
-}
-</style>


edwh:
Ok.  What's the general lesson to bear in mind when writing components?

JayyajGH:
I'm not sure if there is any big learn from this.  It might even have been me that put this class there in the first place!  

In this case the Autocomplete component had styles set up for "autocomplete ul" (As a side note we should never be targeting elements but this is because we are overriding bootstrap which is a whole other argument).  The Postcode component that contained the autocomplete component overrode the styles for "autocomplete ul". 

I found this confusing as I saw the styles defined in autocomplete but couldn't see why in the browser they weren't being used.  Because they were in a different component they were hard(er) to find.  I would suggest that all the styles that are used in the lowest component they are used in - unless they are generic. We were already passing in the class name anyway.

The more I type, the more I think this is just a weird setup and there isn't anything to learn.  The whole idea of overriding styles in a component using v-deep is horrible. I think if we are doing things like that then we have to accept that some code will be horrible.  Ideally the component provider should provide enough flexibility in props etc. to achieve what you want.

edwh:
Ok.  I feel I should write something longer than that, but I think you're right that doing horrible things leads to things being horrible.

I've done a horrible thing with CSS in ModGroupMap to get maps working quickly.  But you'd need to be a mod on a test group to sort that out.

JayyajGH:
A lot of my CSS style has come from a guy called Harry Roberts who has done some really cool stuff.  I'll try and find a link to the video of his I really liked.  One of his points is to have a file called shame.css which contains the stuff you know is bad.  I think it's perfectly acceptable to write bad CSS sometimes but call it out.  Make it obvious.  Everyone has time pressures etc. or just need to get stuff working sometimes.  If you put big comments with ****** SHAME ****** or something then it just makes it clear and you can revisit it when you have time.

edwh:
I'm a big fan of knowing when you're being a bit crap.  If you're never 
crap, you never get anything done.  If you pretend you're never crap, 
you're kidding yourself.

JayyajGH:
Yeah bad code is sometimes the only way to get stuff done.

JayyajGH:
I really like the Scrum adage "make it work and then make it right".  Basically a badly written feature can start making money as soon as it is released.  Get the benefit asap and then make it tidier later.  Which with suitable automated testing should be easy and risk free.

edwh:
To a point.  It's also a recipe for "make it work and then....make 
something else work".  I've not worked on an agile project which didn't 
have ever-increasing tech debt.

JayyajGH:
So true.  You need to make sure the product owner and scrum master are given complete authority to control the work so technical debt doesn't build up.  Which rarely happens.

PR #292, opened 2020-07-01
Disassociate the semantic headers from the visual aspect

JayyajGH:
A first go at making some of the pages document structure correct.

The browse, volunteer, volunteerings, event and events pages should now have the correct structure.  I've added some headers that are screen reader only so hidden to sighted users.  The bootstrap card already provides the feature to set the element used for the titles and so I've just used that to pass in the heading level.

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -137,6 +137,22 @@ h1, h2, h3, h4 {
   color: $colour-header !important;
 }
 
+.header--size1 {
+  font-size: 2.5rem;
+}
+
+.header--size2 {
+  font-size: 2rem;
+}
+
+.header--size3 {
+  font-size: 1.75rem;
+}
+
+.header--size4 {
+  font-size: 1.5rem;
+}
+
 div[variant="success"] .card-header {
   color: $colour-success-fg;
   background-color: $color-green--light;
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -153,6 +153,10 @@ h1, h2, h3, h4 {
   font-size: 1.5rem;
 }
 
+.header--size5 {
+  font-size: 1.25rem;
+}
+
 div[variant="success"] .card-header {
   color: $colour-success-fg;
   background-color: $color-green--light;
--- a/components/CommunityEvent.vue
+++ b/components/CommunityEvent.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <b-card variant="success" no-body>
-      <b-card-title class="bg-info px-2 mb-0 pt-2 pb-2 text-truncate d-flex justify-content-between">
+      <b-card-title class="bg-info px-2 mb-0 pt-2 pb-2 text-truncate d-flex justify-content-between header--size4" :title-tag="titleTag">
         <nuxt-link :to="'/communityevent/' + item.id" class="event__link">
           {{ item.title }}
         </nuxt-link>
@@ -126,6 +126,11 @@ export default {
     item: {
       type: Object,
       required: true
+    },
+    titleTag: {
+      type: String,
+      required: false,
+      default: 'h3'
     }
   },
   data: function() {
--- a/components/CommunityFeature.vue
+++ b/components/CommunityFeature.vue
@@ -4,11 +4,11 @@
       <b-card-body class="p-0">
         <div class="p-1">
           <div class="d-flex align-items-start justify-content-between">
-            <h4>
+            <h2 class="header--size4">
               <nuxt-link :to="link" class="title__link">
                 <v-icon :name="iconName" scale="2" /> {{ title }}
               </nuxt-link>
-            </h4>
+            </h2>
             <b-btn variant="white" :aria-label="addButtonLabel" @click="showEventModal">
               <v-icon name="plus" /> Add
             </b-btn>
--- a/components/ExploreMap.vue
+++ b/components/ExploreMap.vue
@@ -7,9 +7,9 @@
         </div>
         <div v-else>
           <h1>Explore Freegle communities across the UK!</h1>
-          <h5 v-if="groupCount">
+          <p v-if="groupCount" class="community__heading">
             There are {{ groupCount }} lovely communities of freeglers across the UK. Shall we see what they're up to?
-          </h5>
+          </p>
         </div>
       </b-col>
     </b-row>
@@ -32,9 +32,9 @@
     </b-row>
     <b-row v-if="!region && regions.length" class="m-0">
       <b-col cols="12" lg="6" offset-lg="3" class="mt-2">
-        <h5 class="text-center">
+        <h2 class="text-center header--size5 community__text">
           Choose a region:
-        </h5>
+        </h2>
         <b-list-group horizontal class="flex flex-wrap justify-content-center">
           <b-list-group-item v-for="r in regions" :key="r" class="p-0 mt-2 ml-2 mr-2">
             <b-btn variant="white" :to="'/explore/region/' + r">
@@ -63,6 +63,9 @@
         </client-only>
       </b-col>
     </b-row>
+    <h2 class="sr-only">
+      List of communities
+    </h2>
     <b-row class="m-0">
       <b-col v-if="groupsInBounds.length" cols="12" lg="6" offset-lg="3" class="mt-4">
         <b-card header-bg-variant="success" header-text-variant="white" header="Here's a list of communities:">
@@ -121,9 +124,9 @@
             </client-only>
           </b-card-body>
         </b-card>
-        <h5 class="text-center mt-2">
+        <p class="text-center mt-2 header--size5 community__heading community__text">
           If there's no community for your area, would you like to start one? <a href="mailto:newgroups@ilovefreegle.org">Mail us!</a>
-        </h5>
+        </p>
         <p class="text-center">
           You can also look at our <nuxt-link to="/livemap">
             live map
@@ -370,3 +373,16 @@ export default {
   }
 }
 </script>
+
+<style>
+.community__heading {
+  font-size: 1.25rem;
+  font-weight: 500;
+  line-height: 1.2;
+}
+
+.community__text {
+  /* Need to override the h2 as it has higher specificity */
+  color: #212529 !important;
+}
+</style>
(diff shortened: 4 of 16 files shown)


edwh:
I'm in the middle of other stuff, and it looks like you're cracking on with this.  Give me a shout when you need it reviewing in case I get distracted.

JayyajGH:
This should now be the bulk of the easy changes.  There are still issues with chats and modal popups etc. but that will be much more complex to resolve so it's probably best to get this set moving as it's an improvement.

JayyajGH:
So just feel free to look at it when you get a spare 5 mins!

edwh:
What's the rationale for underscores and dashes in class names?

For example, you have "community__heading", and "header--size4".  

There's probably single dash, single underscore, camel case elsewhere in the code.  

Should we have a convention?

JayyajGH:
Sorry I automatically flip into BEM convention when writing CSS.  Block__element--modifier.  It fits in really well with atomic css and utility first css etc. So for buttons for example would be:

.button for the base generic button stuff
.button--small, button--medium, button--large for the sizing
.button--primary, button--secondary, button--tertiary for the button type
.button__label, .button__icon etc. to refer to part of the button


https://css-tricks.com/bem-101/

JayyajGH:
I think this style works really well with components as the block can be your component and the stuff that makes up the component are your elements.

edwh:
Thank you - now I know.

The stuff itself looks fine, and better.  Ok to merge or are you still 
doing it?

JayyajGH:
https://www.youtube.com/watch?v=wFn5nel3j6w
https://speakerdeck.com/csswizardry/css-for-software-engineers-for-css-developers

This video is worth a worth.  I've seen Harry Roberts talk many times and I think he is probably the person that has taught me the most about CSS.  Not that BEM is his idea but this talk mentions using it in the atomic CSS style.  His general idea is CSS should be treated like any other programming language.  Single responsibility principle and all that. Classes should do one thing.  This is where the CSS utility frameworks like Tailwind come in which loads of people use with Vue.

JayyajGH:
Happy to change the classes though if you would rather it fitted it better with the other CSS...

JayyajGH:
This is complete for now.  There are other changes but they will be the 20% of the 80/20 rule!

edwh:
It's a little inconcsistent with bootstrap, but it seems like a good 
rationale, so stick with it.

JayyajGH:
If you haven't seen it before then check out tailwind.  I'm still not sure how i feel about utility first css but it's getting more and more popular.  It's basically like bootstrap but without the opinionated components.

edwh:
I like opinionated :-).  Someone else's considered choices are often better than my own floundering in a sea of options.

I'll check out the links, ta.

JayyajGH:
I think my concern about these kind of frameworks is they almost discourage you from having consistent styles.  If you set up your component and styleguide well then you don't really need a lot of this stuff.  You probably shouldn't be randomly adding padding to an element etc.  Otherwise someone will use a component and add 10px and someone else will add 20px and you lose the consistency.

JayyajGH:
I think my last comment sort of agreed with your last statement :-)

JayyajGH:
When i add a button to a page then i don't want to worry about how much padding to give it and how to lay it out.  I want whatever the button component designer set it up to have. I want a medium sized primary button.  Full stop.  That button knows how much padding to have and how big the font is etc.

edwh:
I agree about the padding kind of thing.  All that nonsense I did with 
b-row should have told me that I was doing the wrong thing by using 
those components.

JayyajGH:
Hindsight is a great thing!  In reality designing up front never works either as people change their minds.

edwh:
Well, nothing actually works, does it?

edwh:
Very interesting talk indeed.

I think there's a key distinction he doesn't bring out, though.

If you're structuring your "stuff" (code, CSS, whatever) for an expert 
in that area who will be working on the internals, then you want to do 
all the kinds of stuff he talks about.  It makes your stuff better, 
makes it easier to maintain, etc.  He's quite right that these 
principles work in the CSS world too.  You could go further, though.  He 
is advocating are a whole bunch of principles for people to remember and 
apply; the lesson from the way other programming languages have 
developed is that you embody those principles into languages.  It would 
be a nice world where CSS was reduced to the status of assembly 
language, and you worked in higher level languages.  We're moving in 
that direction, but it seems to be slow in CSS, whereas if you look at 
the way raw JS has evolved into jQuery into React/Vue that side is 
moving faster.

However if you're providing your "stuff" to someone else, then it's a 
completely different situation.  It is about interface design (not 
visual interface, conceptual interface).  You could write an interface 
which exposes a zillion different options, and leave it up to the user 
to combine those in sensible and good combinations.  But most of those 
combinations will not be sensible and good.

"Gives you incredible opportunity and flexibility" sounds good, but it's 
passing the buck.  The example of Subway being able to make 6.4 billion 
different sandwiches shows that you've offloaded an infeasible amount of 
choice onto the user.  You'll get a better sandwich by going to a 
sandwich shop where someone who knows a lot about sandwiches has chosen 
a selection of good sandwich combinations.

The ways in which I've used low-level Bootstrap utility classes to 
construct higher level clickable components in ways that break 
accessibility is a good example of how if you present someone with a 
toolkit of components which all have a single clear purpose then they 
can combine them in dumb ways.

Restricting choice at the interface level means you won't make very bad 
choices.  You won't make very good choices either.  But if you're not an 
expert in that area, you're much more likely to make old bad choices 
than inspired new good choices.  Less choice means a better average 
outcome.

This is why Bootstrap (despite the above example) is great for people 
like me who are not CSS experts, because it provides an actual interface 
that means an idiot can't go too far wrong.  And tailwind is great for 
people who are CSS experts who want to use it to build an actual 
interface.  Opinionated is good at the interface level, and bad 
internally.

I'd argue that this distinction between interface design and internal 
design is, ironically, one of the lessons of software development which 
has not yet (IMNSHO) been properly absorbed by CSS folk.  Consequently 
CSS folk tend to impose internal complexity on external interfaces and 
think that it's freedom, when it's more likely to be chaos.

JayyajGH:
Really interesting discussion.  Yeah I'd completely agree with you on many points.  I take your main point to be that consumers of components shouldn't have too much control.  That should be left to the component (or system) designer.

The interfaces (visual) and interfaces (conceptual) should ideally be designed up front and set in (soft) stone.
A consumer of components shouldn't have to worry about colours and padding and positioning.  That should be defined at some higher level.  It's why I'm still
not entirely comfortable with the the utility driven CSS frameworks like Tailwind and to some extend Bootstrap too.  They give the consumer too much control and choice.
It starts off good - give me a variant success button. Great.  But then you can tweak the padding and the spacing and the colours etc.  It seems to miss the whole point.

I suppose I didn't take it that he was saying you should provide loads of different options to a consumer.  Buttons for example.  You might start with button--primary, 
button--secondary and button--tertiary.  I would probably keep those in a _buttons.scss file or something similiar.  A button component consumer wouldn't just create
buttons themselves.  This would then be up to that component designer to create a new button type if required like button--ghost.  It's great for the button designer to be able to create lots of options quickly and easily.  I guess the problem with CSS is that then presents the opportunity for the consumer to do it too. The fact that it is easy to override pretty much anything in CSS is almost it's downfall.  It's just a verbal contract.

I think if you use Bootstrap out of the box without really knowing CSS it's great.  I don't think you fit into that box though as (even though you say you don't) you actually know CSS pretty well.  That means you can easily tweak bits which I suppose starts to break the interface defined by Bootstrap.

JayyajGH:
I think I might have mentioned this before but a few years back I tried to create a CSS "starter pack" which tried to address what we're talking about.  It set in stone font sizes, colours, padding for everything you would want to use.  You want to add a heading?  Well if you want a size 2 heading then you must have it with 15px padding under it.  It was meant to be that the consumer wouldn't really have to write much CSS themselves.  Although if they would still be technically able to override whatever they wanted.  The idea was that when you started a project you would add this package in, tweak the defaults to be what you wanted and then it just worked.

It's here if you were bored and wanted to take a look at any point:
https://github.com/JayyajGH/Sass-Boilerplate

JayyajGH:
Styleguide driven development also goes some way to address this. But once again it's based on trust rather than anything secure.

edwh:
I think he's mostly talking about how to write maintainable internals code here, so I'm criticising him for not giving a different talk, which is a bit silly.  

I think a lot of issues with CSS come from the fact that (in code terms) it readily enables spaghetti code where everyone is using global variables.  People are always going stand on each other's toes, and then you evolve complex conventions and coding standards to work around that.  

Really that is always going to be a problem as long as those kind of concepts like public and private aren't built into the language.  The CSS scoping trick in Vue components is a nice and useful hack, but those kind of features need to be compile-time errors in the language rather than optional best practice and bugs at run-time.  

But of course, that's why it's successful.  Sloppy languages like JS, PHP, CSS are great at conquering an evolving and open landscape, and then later on you tidy them up.  JS and PHP are getting there, CSS is still in a world where there are sheriffs trying to tame the Wild West.

I agree that in terms of CSS I fit with the "a little knowledge" cliche.  I know enough to pervert Bootstrap, but not enough to not need it :-).

JayyajGH:
Another point we haven't talked about is that I think people see CSS as being "just a styling thing" and so people just knock up a complete mess because it's not seen as a "real" language.  The same as any programming language, if you aren't good at it then you create a mess in the same way as people would with C#.  In my head this is changing but that might just be because of the people that I follow on twitter etc.  It gives you a false sense of the world.

edwh:
Yes, I think that's very true.  CSS is perceived as incidental - either 
your focus is the code, or your focus is the visual stuff, but you don't 
feel you need to focus on CSS.

That reminds me of this "Javascript: The Good Parts" talk, which is 
quite old, but I remember being excellent.  And funny:

"It's the only language, that I'm aware of, that people feel that they 
don't need to learn it before they start using it."

https://www.youtube.com/watch?v=hQVTIJBZook

JayyajGH:
I've never seen that talk but I have read the book.  It's very good.  I'll watch that later.  I find Crockford a bit like the author Richard Dawkins.  Loads of good points but they come across too extreme.

JayyajGH:
The pluralsite course called the same thing is also very good.

PR #294, opened 2020-07-13
Feature/refactor message

JayyajGH:
This is the refactor of the message component to use Grid so it doesn't duplicate the HTML for desktop/mobile

The HTML is now much cleaner and the differences between desktop and mobile styles are dealt with where they should be in the CSS.

I've tested this on Chrome, Edge, Firefox, Safari and IE11.  IE11 looks ok but the element order is slightly messed up.  We can add IE prefixes to make this work if you really don't like how it looks.  On mobile I've tried it on a Galaxy S20, a Pixel 2 and an iPhone 11.  I'm sure you will but it would be good for you to give this a proper test in case I'm missing some functionality or pages.

Firstly how important is it to not display the ID in the message history for mobile?  This works in this PR (using the sm function you have previously written) but at the moment it doesn't feel very performant as it will be called for every message.
The options would be:

1. Just display it for mobile as well
2. Cache the figure either simply (once at the beginning which could mean it becomes out of date) or in a more complex way with some kind of watch (not entirely sure how that would work at the moment).

The other point is i'm not going to be near a computer from Wednesday evening until the following Thursday so you might want to hold off merging this until I'm back.

### DIFF
--- a/components/Message.vue
+++ b/components/Message.vue
@@ -2,150 +2,67 @@
   <div>
     <span ref="breakpoint" class="d-inline d-sm-none" />
     <b-card class="p-0 mb-1" variant="success">
-      <b-card-header :class="'pl-2 pr-2 clearfix' + (ispromised ? ' promisedfade' : '')">
-        <b-card-title class="msgsubj mb-0 d-block d-sm-none header--size4" title-tag="h3">
-          <div>
-            <div>
+      <b-card-header :class="'pl-2 pr-2 clearfix card-header' + (ispromised ? ' promisedfade' : '')">
+        <b-card-title class="msgsubj mb-0 header--size4 card-header__title" title-tag="h3">
+          <Highlighter
+            v-if="matchedon"
+            :search-words="[matchedon.word]"
+            :text-to-highlight="eSubject"
+            highlight-class-name="highlight"
+            auto-escape
+          />
+          <span v-else>
+            {{ eSubject }}
+          </span>
+        </b-card-title>
+        <MessageHistory :message="$props" class="card-header__history" :display-message-link="sm()" />
+        <div flex-grow-1 class="small card-header__description">
+          <div v-if="eSnippet && eSnippet !== 'null' && !expanded">
+            <b class="snippet black">
               <Highlighter
                 v-if="matchedon"
                 :search-words="[matchedon.word]"
-                :text-to-highlight="eSubject"
+                :text-to-highlight="eSnippet"
                 highlight-class-name="highlight"
                 auto-escape
               />
-              <span v-else>
-                {{ eSubject }}
-              </span>
-            </div>
-            <MessageHistory :message="$props" />
+              <span v-else>{{ eSnippet }}</span>
+              ...
+            </b>
           </div>
-          <div v-if="attachments && attachments.length > 0" class="d-block mt-1 d-sm-none clickme position-relative" @click="showPhotos">
-            <b-badge v-if="attachments.length > 1" class="photobadge" variant="primary">
-              {{ attachments.length }} <v-icon name="camera" />
-            </b-badge>
-            <b-img-lazy
-              rounded
-              fluid-grow
-              class="attachment p-0 square"
-              generator-unable-to-provide-required-alt=""
-              title="Item picture"
-              :src="attachments[0].paththumb"
-            />
+          <div v-if="(!eSnippet || eSnippet === 'null') && !expanded">
+            <i>There's no description.</i>
           </div>
-          <div class="small">
-            <div v-if="eSnippet && eSnippet !== 'null' && !expanded">
-              <b class="snippet black">
-                <Highlighter
-                  v-if="matchedon"
-                  :search-words="[matchedon.word]"
-                  :text-to-highlight="eSnippet"
-                  highlight-class-name="highlight"
-                  auto-escape
-                />
-                <span v-else>{{ eSnippet }}</span>
-                ...
-              </b>
-            </div>
-            <div v-if="(!eSnippet || eSnippet === 'null') && !expanded">
-              <i>There's no description.</i>
-            </div>
-            <b-button v-if="!expanded" variant="white" class="mt-1" @click="expand">
-              See details and reply <v-icon name="angle-double-right" />
-            </b-button>
-          </div>
-        </b-card-title>
-        <b-card-title class="msgsubj mb-0 d-none d-sm-block header--size4" title-tag="h3">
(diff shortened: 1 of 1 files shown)


edwh:
I'm away 15th to 20th and busy tomorrow, so I will leave this one till we're both back.

Should be an improvement once we're happy with it, plus I will brush up on grid.

JayyajGH:
ok cool.  This is pretty much all I've ever used grid for.  I'm certainly no expert with it.  It's really just the basics of Grid so should be really simple to understand.  I think the thing to know about it is that it is basically just a 2D Grid in which you lay out your elements.

JayyajGH:
Looking nicer...

image

edwh:
I checked it out and had a play.  I can't make it break, but I don't 
think I understand grid well enough to review the changes well.

This is fairly complex CSS, at least for me.  At the moment, either you 
understand enough for it to be obvious, or you bang your head against it 
for quite a while.  Could you try adding some comments?

But there are points where it looks a bit messy.  This isn't down to 
your change, I don't think.  The "Close post..share" line just doesn't 
feel right.  The photo feels like it has space to be bigger, but I can't 
see how.  It just looks a bit lame.



I've wondered about a side by side layout card for messages, so that we 
get a view more like the one on mobile, and then we have as many of 
those side by side as makes sense at that width.  Don't know what you 
think about that?


JayyajGH:
So I was meant to fly back today but we decided to bin off the flight and travel further south to Rome and Naples. Planning on heading back home next Tuesday now so I'll have a proper read of your comments then and have a chat.

edwh:
No rush. Have fun.

JasonHarrisEIS:
Finally home.  And back to work....

I didn't add comments as I thought the comments would be too simple.  I'll try and write a brief description of what I've done here to see if that is enough.  What I've done here is pretty much the same thing I did on the GroupHeader.

There are loads of different ways you can layout things in grid.  I think maybe three different styles but they all achieve the same thing. I used this style here as it's probably the most explicit.

The only real grid stuff I've done is organising the layout of the rows and columns.  So something like "grid-row: 1/2".  This basically means that this content starts from row 1 and continues to row 2. If you wanted the element to take up more rows then you would change it to "grid-row: 1/4" for example.

Another way to write this is 
grid-row-start: 1;
grid-row-end: 2

The "/" is just shorthand but I think it makes it easier to read and see the layout at a glance.

The grid-template-rows and grid-template-columns lines are just explicitly stating how the rows and columns should be laid out.  Otherwise it will just decide for itself.  There are a pile of ways of stating the size of rows and columns but max-content will basically expand to fill the content.  The size of this here is defined by having explicit widths on the image etc. which I haven't changed.

Does that help at all?  Is that enough?

The site I always go to for grid help is the same one I use for flexbox:
https://css-tricks.com/snippets/css/complete-guide-grid/

These games are really good for an initial intro to grid too.  I've done the flexbox one but there is a grid one too.
https://codepip.com/games/grid-garden/

JasonHarrisEIS:
The close post/share line is very slightly different in my version but only very slightly.  I do agree that it feels a bit odd.  The problem with making the image bigger is that with the same dimensions there will be a lot of empty space.  Yeah maybe the layout needs a complete rethink.  I'll take a look at some other similar sites to see what they do.

JayyajGH:
Do we need the complexity of expanding the items in-place?  Looking at something like ebay and it's just an image and a title.  If you want to see more then you click on it.  Only then do you get the option to view the pictures and interact with it.  Amazon works exactly the same way. So does Gumtree.

edwh:
If we did that we'd need the back button to take you to the same place 
in the infinite scroll.  That's possible, and I think there's an issue 
somewhere, but it's complex.

JayyajGH:
Ah yes, the pain of infinite scroll.  Was there any specific reason for going for infinite scroll over pagination? (A theoretical question more than suggesting that we change everything!).  Anything with the back button is usually complex. I've spent hours trying to fix problems with the browser bfcache in the past.

edwh:
If in doubt, follow Facebook.  Pagination is tedious - you have to click 
on a fiddly little thing rather than just swipe/scroll.

You can do it, e.g. using IntersectionObserver, but it needs a bit of 
work.

JayyajGH:
I guess there are advantages and disadvantages of both.  Infinite scroll causes problems with accessibility and SEO but can also be much quicker for the user.  I'm on the fence with my opinion on it.

Shall we take the improvements to the message design out of this conversation and add a new issue for it? Whatever we do to it would be easier with the Grid changes as it removes the duplication.  Obviously we still need to address the initial comments you made...

edwh:
I agree, it's a separate issue.

JayyajGH:
We're building up a good collection of things we don't want to talk about! :-)

edwh:
Thanks.  I need to spend a bit of time on it, I think.  Good pointers.

JayyajGH:
ok cool.  Just let me know if you need any more info or want me to change anything.

When creating something like this I usually draw out the content on paper and then draw numbered grid lines on it.  It then becomes immediately obvious.  I do one diagram for each breakpoint.

JayyajGH:
Does this all make sense now then?

edwh:
I can't claim I've got my head round grid.  But that's my issue.

JayyajGH:
Would it help if I drew a quick diagram of this page and the grid it uses?  Happy to assist you getting your head around it.  The grid garden game is definitely worth a go though if you have a spare 30mins.

edwh:
I'll try the game.  No need to waste your time.

JayyajGH:
I couldn't help myself...

edwh:
Ha!  Will have a look.

edwh:
Thanks.  I had a read and a play of the game, so I reckon I now know 
enough to have a crack at using grid next time.

JayyajGH:
The other confusing thing that seems to confuse everyone is knowing when to use grid and when to use flexbox.

edwh:
I'm sure.  I googled, and found 
https://css-tricks.com/css-grid-replace-flexbox/ .  That mentioned 
Rachel Andrew.  You might not know that she was involved right at the 
start of Freegle, and helped us get a site up and running in the heady 
early days, using her CMS Perch.  I was entertained to find that.

JayyajGH:
That's really cool. I've seen her talk a load of times and she lives in Bristol. She is pretty much the one authority on flexbox and grid.

JayyajGH:
And weirdly I've done a lot of work on a codebase that was written by the guy that wrote that article.

PR #300, opened 2020-08-05
Style the notification dropdown toggle to be the correct colour

JayyajGH:


edwh:
I'm sure this works, but the way you've done it is quite specific to this case.  Does that mean that other toggles will be wrong, or is it because of the other styling we do in this case?

JayyajGH:
I usually hate anything this specific but in this case I tried to make it specific as possible so as not to affect anything else that I didn't know about. This is mostly due to my lack of knowledge of bootstrap components I think.  Styling these hidden entities always makes me nervous.

Is this toggle style used anywhere else?  i.e. a bootstrap dropdown component which creates a toggle button. In this case it doesn't really fit into our regular list of buttons.  I thought about taking this further and creating a common style for all the nav items which I might do next.

Previously there was a variant property on the dropdown which bootstrap carries through to the contained button.  There is another button contained within the dropdown but it is within our control and styled independently so that wasn't affected.  The ones used on chat appears to be slightly different too.

JayyajGH:
I think this probably broke when I recently consolidated all of the button styles in one place.  It started acting as a regular primary button which it isn't and probably should never have had the variant="primary" attribute on it.

edwh:
I can't think of other ones.  There are a couple of other dropdowns, 
ChatHeader and NewsThread, but they don't contain buttons, just menu 
options.

Due diligence done.  I'll merge it - ta.

JayyajGH:
Actually I might have changed my mind.  Just discovered there is a bootstrap variant transparent which is used in a few other places.  The colour is then overriden to white in other menus.  I'll create a new PR and change it to this style to keep things consistent.

JayyajGH:
There are then global settings in global.scss which sort out the off-white colour

edwh:
I'm a cardboard dog, you know.

https://www.sjbaker.org/humor/cardboard_dog.html

JayyajGH:
Never heard the term before but someone I used to work with had a doll on her desk for a similar purpose.

PR #302, opened 2020-08-05
Remove the bootstrap rows to fix the mobile sizing issue.

JayyajGH:
Fix the mobile layout by replacing the bootstrap rows/cols with flexbox

### DIFF
--- a/components/CommunityEvent.vue
+++ b/components/CommunityEvent.vue
@@ -51,58 +51,54 @@
             Posted on {{ item.groups[0].namedisplay }}
           </div>
         </div>
-        <div v-else>
-          <b-row>
-            <b-col cols="12" :md="item.photo ? 6 : 12">
-              <div v-if="item.earliestDate" class="media">
-                <div class="media-left">
-                  <div class="media-object pl-1 text-muted">
-                    <v-icon name="clock" class="fa-fw" />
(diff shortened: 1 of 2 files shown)


JayyajGH:
For now I have just placed the image underneath the text and left-aligned.  It was in a bit of a weird place before.  This is least fixed the left/right shimmy on mobile.

edwh:
Same issue presumably applies to CommunityEvent - can you do that one 
too as part of this?

JayyajGH:
Was gonna look at that one next. Yup can do it as part of this (as long as it's not too weird). I thought I'd added a community event when I added the volunteer op.  Can you see if it's still pending moderation?

JayyajGH:
Looks like exactly the same issue with that one so should be nice and easy

edwh:
Nope, not there.

Shall I make you a mod on the Playground group, then you can approve 
your own?

JayyajGH:
Yeah if that's easy enough to do.  How do I get to the mod area?

edwh:
Done.  Access via modtools.org or localhost:3000/modtools.

JayyajGH:
I'll try and not let the power go to my head. :-)

JayyajGH:
That's events sorted now too.

PR #303, opened 2020-08-06
Remove the bootstrap rows/cols to fix the mobile shimmy

JayyajGH:
Fix the mobile row/col issue shimmy on the give page.

### DIFF
--- a/pages/find/whatnext.vue
+++ b/pages/find/whatnext.vue
@@ -18,32 +18,31 @@
         </p>
       </b-col>
     </b-row>
-    <b-row class="text-center">
-      <b-col cols="12" class="mt-2">
+    <div class="text-center">
+      <div class="mt-2">
         <b-btn size="lg" variant="primary" to="/find/whatisit">
           <v-icon name="gift" />&nbsp;Find something else
         </b-btn>
-      </b-col>
-      <b-col cols="12" class="mt-2">
+      </div>
+      <div class="mt-2">
         <b-btn size="lg" variant="secondary" to="/give/whatisit">
           <v-icon name="search" />&nbsp;Give something
         </b-btn>
-      </b-col>
-      <b-col cols="12" class="mt-2">
+      </div>
+      <div class="mt-2">
         <b-btn variant="white" size="lg" @click="availability">
           <v-icon name="calendar-alt" /> Edit Availability
         </b-btn>
-      </b-col>
-    </b-row>
-    <b-row>
-      <b-col class="text-center mt-4">
-        <nuxt-link to="/">
-          <b-btn variant="white" size="lg">
-            Continue to Home Page <v-icon name="angle-double-right" />
-          </b-btn>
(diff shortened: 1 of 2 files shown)


edwh:
I haven't checked (off to the vet) but may apply to /find/whatnext too?

JayyajGH:
Oh yeah it does.  We really need to create a common component for these...  I'll sort that one as part of this PR.

Good luck at the vets!

JayyajGH:
Fixed on find as well as give now

edwh:
Shall I take this now?  Shared component could come later, as there may 
be wrinkles.



JayyajGH:
Yup now is good.  Unless you can think of anything weird on this page...  I've only tested this in Chrome.

edwh:
The more I learn, the more I feel I want everything that's part of bootstrap apart from the layout.

JayyajGH:
That's an option though.  You can just bring in the stuff you need can't you?  Sounds like you want something halfway between tailwind and bootstrap.  I think now we have flexbox and grid as part of CSS, it's probably easier to use than the bootstrap implementation.  I think the JS frameworks will go the same way.  Same as things like Sass.  They are slowly becoming obsolete as the basic tools are getting better.

edwh:
Yes, you can do that.  And when I first used bootstrap it was from a 
position of knowing very little CSS and wanting a way to produce a 
responsive layout without really understanding what I was doing.  It did 
the job for that, it's not bootstrap's fault if I've got marginally more 
knowledgeable.

JayyajGH:
I guess that's the problem with any project that lasts more than about a year.  Things change and become out of date very quickly.  I've been trying to resolve an issue with C++ COM this morning.  Freegle will be fine. :-)  Freegle would probably still be in development if you had manually crafted everything!

edwh:
https://en.wikipedia.org/wiki/Wittgenstein%27s_ladder

JayyajGH:
I've not heard that before but it's so true!  I doubt anyone would do a project the same way once they have finished it.  And no one every has time to go back and throw away the ladder.

edwh:
Probably if we spoke to people who made ladders they'd do it all 
differently if they had the chance.

JayyajGH:
I was struggling putting up some plasterboard last night as I had nowhere to put my drill and screws when I was stood on it.  This morning I saw a ladder with a little holder for all those things.  Ladder creators are what we aspire to be. :-)

edwh:
I tell you something that's due a v2.  Wheels.

JayyajGH:
We've had enough time...

edwh:
https://www.quora.com/How-many-alternatives-to-the-wheel-are-there

JayyajGH:
The screw drive one is really interesting.  Not sure how efficient that would be though!

edwh:
It would be important to consider the direction of the thread on each 
side carefully.

PR #304, opened 2020-08-09
Convert A mailto into ExternalLink so they work in ios app

chriscant:
Updated a load of A mailto links to use ExternalLink so they work in iOS app

### DIFF
--- a/components/CommunityEventModal.vue
+++ b/components/CommunityEventModal.vue
@@ -96,9 +96,7 @@
               Contact email
             </b-col>
             <b-col cols="8" md="9">
-              <ExternalLink :href="'mailto:' + event.contactemail">
-                {{ event.contactemail }}
-              </ExternalLink>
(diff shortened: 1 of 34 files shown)


edwh:
Before I review in detail, I notice that in e.g. CommunityEventModal the change will result in whitespace inside the <a>.  I have cruelly forced you to avoid that in an earlier fix.  Can you do so in this?

Slightly more generally, I wonder whether for ExternalLink we should use a prop rather than a slot?  That would allow us to handle the spacing issue inside that component.

chriscant:
Ah yes, 
My editor automatically formatted it, which caused the probem.
I'll check the whitespace now, and make a note about converting to a prop.

edwh:
Thanks.  Apologies in advance for when I forget to use it.  I really need to sort out some project lint rules to flag up stuff like this.

chriscant:
Thanks. There was one mailto on a button which I didn't do - to check our iOS app testers

PR #305, opened 2020-08-12
Allow the text to wrap so it doesn't push out the screen on mobile

JayyajGH:
I haven't given this much of a test yet but thought i'd raise a PR for discussion.

This allow the "2 days ago on FreeglePlayground #1234565" bit to wrap if necessary.  It also allows the "userid last sent" bit of replies to wrap if the username is too long.

See the slack thread for more info...  There are screenshots of the problem attached there.

Are you happy with this wrapping when it doesn't fit or would you rather truncate it?

### DIFF
--- a/components/MyMessage.vue
+++ b/components/MyMessage.vue
@@ -23,7 +23,7 @@
                     <v-icon name="exclamation-triangle" scale="2" />
                   </span>
                 </h3>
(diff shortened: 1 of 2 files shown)


edwh:
Wrapping both is fine.

I have seen some people with stupid long usernames, but that should be 
handled outside this PR.

JayyajGH:
OK I'll give this a bit more of a test this afternoon then

PR #308, opened 2020-08-28
Feature/map animation

JayyajGH:
Re. visualise map animation stuff discussion on slack....

How about something like this?  A bit of fade in and a bit of bounce.  Excuse the coding as this is just knocked up for a demo.  Just checking if this is the sort of thing you are after

### DIFF
--- a/api/BaseAPI.js
+++ b/api/BaseAPI.js
@@ -59,7 +59,7 @@ export default class BaseAPI {
       })
       ;({ status, data } = ret)
 
-      if (!status || !data) {
+      if (!status || !data || !('ret' in data)) {
         // We're investigating some cases, with some evidence of it happening in page unload, where we don't go
         // through the exception handler, but we end up with nothing in status/data.  Throw an exception to trigger
         // a single retry.
--- a/api/NoticeboardAPI.js
+++ b/api/NoticeboardAPI.js
@@ -14,6 +14,10 @@ export default class NoticeboardAPI extends BaseAPI {
     return this.$patch('/noticeboard', data)
   }
 
+  action(data) {
+    return this.$post('/noticeboard', data)
+  }
+
   del(id) {
     return this.$del('/noticeboard', { id })
   }
--- a/components/ConfirmModal.vue
+++ b/components/ConfirmModal.vue
@@ -39,7 +39,7 @@ export default {
     }
   },
   methods: {
-    show(type) {
+    show() {
       this.showModal = true
     },
 
--- a/components/GroupHeader.vue
+++ b/components/GroupHeader.vue
@@ -57,9 +57,11 @@
       <h2 class="group-volunteers mb-3">
         You can contact our lovely local volunteers here:
       </h2>
-      <b-button class="mb-3" variant="white" :href="'mailto:' + modsemail">
-        <v-icon name="question-circle" />&nbsp;Contact&nbsp;volunteers
-      </b-button>
+      <ExternalLink :href="'mailto:' + modsemail">
+        <span class="btn btn-white mb-3">
+          <v-icon name="question-circle" />&nbsp;Contact&nbsp;volunteers
+        </span>
+      </ExternalLink>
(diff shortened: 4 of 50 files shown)


edwh:
This is the kind of thing, yes.  Suggestions - doesn't mean you need to 
do them all, just food for thought:
More bounce than squish - not sure that works well with people's faces.
In the version I had ages ago it dropped in from the top, fading in as 
it arrived, then bounced, a bit higher than yours.  That was all horrid 
JS.
Probably a bit bouncier when it first appears.  Maybe a bit faster so 
that it feels more of a sense of excitement than resting pulse?
Ideally not all bouncing at the same.
Maybe a shrink and fade for the removal.
Similarly some stuff on the photo and textboxes, but probably just 
expanding and fading in, and a speech arrow thingy.

But yes, exactly along these lines.  You could spend forever on it, and 
that'd be daft.

JayyajGH:
I'll have a play with some tweaks.  Yeah good point re. the squished faces.  It just felt a bit dead with just bounce so I tried to liven it up a bit.  I'll try what you did originally then and see how that looks.  Not having them all doing the same thing at once might be trickier with this solution.  I guess we could use first-of or whatever the CSS properties are called. Watch this space...

edwh:
We can always delay the add in JS.

JayyajGH:
Yeah true.  I'll try out some different animations first to get something working.  Got a meeting for the rest of the morning so it might be this afternoon...

edwh:
That's ok, out most of the day.

JayyajGH:
A quick tweak before I had off for the morning...

JayyajGH:
Something else we need to be aware of here is the prefers-reduced-motion option.  The rapid moving of the map could definitely cause people issues.

edwh:
That option is new to me. Thanks. 

Probably we would use that option on the landing page to hide the whole of the map. The existing video (which will be below the map) would move up. That would basically then appear the same as currently. 

So I think you can ignore that within the map work.

JayyajGH:
Yeah probably the best idea just not to show it.  It doesn't really work without movement!

JayyajGH:
I've made some tweaks to the bouncyness of the user icon.  See what you think...

edwh:
Yes, that bounciness is nice now.

JayyajGH:
A phrase you don't see very often on a PR

edwh:
True.  I googled to check, but sadly no exact matches.

https://github.com/eBay/NMessenger/issues/19

JayyajGH:
I'll look at the item image next. Might need to tweak some class names as I subsequently realised that VisualiseUser is also used for the item. 
I think we need to be careful were don't go too far here and have stuff bouncing everywhere. I'll try and do something subtle to the image maybe.

edwh:
User fade, too?

The users are the thing to make most bouncy, because that's where the 
enthusiasm should reside.  The image and speech bubbles just need a bit 
of something, so they don't flick into existence and out again in a 
slightly abrupt way.

JayyajGH:
Yeah maybe a slight fade-in would help.  I'll try out some numbers.  In the office today though.

JayyajGH:
@edwh I've now added a slight fade-in to the item image.  See what you think...

edwh:
That's good on the bounciness and image, thank you.

The speech could do with having the arrow thing to make it look like a 
bubble coming out of the icon, and also having some transitions 
(exploding in).

Multiple responses should be offset by a bit to stop them being 
simultaneous, but I can do that in JS later.

JayyajGH:
Speech bubble added...

JayyajGH:
And a bit of a fade-in of the speech bubble.

JayyajGH:
With a lot of stuff on the screen and moving around I'm trying to keep any animations quite subtle.

JayyajGH:
Just found a slight issue.  I'd only fixed one side of the conversation.  I've now done the other but not really tested it fully.  Up to you whether you want to test it or leave it until I get back.

PR #315, opened 2020-10-07
Only the notification button should have white text

JayyajGH:
I've now made the white foreground for dropdown toggles specific to the notification dropdown.  At the time it was only the mobile notifications button that we noticed was the wrong colour so this shouldn't affect anything else.  I've had a quick click around the pages and there isn't anything that stands out.

In the same way that I did a button tidy up so they were all defined in a single place, it might be worth doing the same thing with dropdowns as well.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -904,7 +904,7 @@ svg.fa-icon {
   max-width: 400px;
 }
 
-::v-deep .dropdown-toggle {
+.notiflist ::v-deep .dropdown-toggle {
   color: $color-white;
 }
 </style>


edwh:
Thanks - I couldn't remember what it was that prompted this change.  That looks fine.

What do you have in mind for the dropdowns?  I can see the words, but perhaps I need more coffee.

JayyajGH:
Initially just some analysis into what we have and the different versions.  Then maybe consolidate them all into one place like I did for the buttons.  (I know I wrote the line that did this but) I really don't like the individual overriding for specific cases as it leads to confusion and issues like this.

In the past I've created a bunch of separate Sass files.  e.g. buttons.scss, dropdowns.scss etc. Then pull everything out and make no button styling changes anywhere but in those files.  I think it helps to keep the separation.  I guess this also fits into the issue we currently have outstanding (which I've partially addressed with buttons) about using bootstrap in a better way.

I'll use buttons as an example.  We used the bootstrap variants for buttons but then we overrode parts of them as well.  So a variant-white button was different wherever it was used.  So really we should have created variant-white-1, variant-white-2 etc. (excuse the names!).   

I guess this is just the open/closed principle.  Don't tweak what you have but instead create a new variant.

JayyajGH:
Ages ago I wrote a bit of a Sass starter project that uses this principle.  The general idea I stole from Harry Roberts and his ITCSS methodology although lots of people have suggested similar things just with different names.

https://github.com/JayyajGH/Sass-Boilerplate

edwh:
Yes, I like the idea of different files for different areas.  I'm in favour of that, in a vague hoping-someone-does-it sort of way.

JayyajGH:
Hugo Giraudel always has some good stuff to say on Sass and it's architecture.  They are a good person to follow on twitter.
https://sass-guidelin.es/#architecture

Harry Roberts and loads of other people have suggested similar layouts.  It's slightly different with Vue and the Sass for a component is already contained but I like the idea of splitting out individual areas.  It also leaves the main css file pretty empty and just some basics in it.

PR #316, opened 2020-10-08
Feature/add header to page

JayyajGH:
A small fix I noticed when playing with Wave.  The chat page didn't have a main H1 header.  I've had to add a container to the page to keep Vue and it's one-root-element checking happy.  I think bootstrap rows are supposed to be in containers anyway.

### DIFF
--- a/components/PlaceAutocomplete.vue
+++ b/components/PlaceAutocomplete.vue
@@ -1,27 +1,23 @@
 <template>
-  <div class="d-flex">
-    <autocomplete
-      id="placeautocomplete"
-      ref="autocomplete"
-      v-model="wip"
-      restrict
-      :url="source"
-      param="q"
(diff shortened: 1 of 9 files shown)


edwh:
Do you really need the b-row round the h1?  If so, then is it valid to have it without a b-col?

JayyajGH:
I could get away with not having the row actually.  I seem to remember that you didn't have to have columns or the container but I can't remember.  I only added the container for the single root element - client only doesn't seem to count.
So yes, I think I'll remove the row.

JayyajGH:
Done...

edwh:
I've got loads with just a top-level div; I'm not sure b-container adds 
that much, really.  But since you opened my eyes to the wonder of flex 
and grid I'm trying to get away from using rows/columns where possible.

JayyajGH:
Yeah the container is probably not strictly necessary.  It just ends up being a div with width 100% anyway.  So basically a regular div.

Ah grid and flex... :-)

PR #319, opened 2020-10-08
And an associated label to the autocomplete input

JayyajGH:
All input fields should have an associated label.  This input had a label but it wasn't connected to the input field.

With the label and input in different components it was awkward to associate them together.  

One option would be to pass the input id into the component and use it in both the parent and child but that felt messy.  It felt right to have the label and the input together.

The awkwardness here then comes from the normal label and the screenreader label.  The two options here are pass in a couple of props or to use a slot.  A slot felt a bit overkill.  The only problem with the solution I've gone for is it's not that easy to validate that someone will pass in at least one of the two.  I suppose another way to look at it is that there is no validation on a regular input field so we are not making anything worse than the norm.

Another issue is that autocomplete isn't accessible but that's another story...

### DIFF
--- a/components/Autocomplete.vue
+++ b/components/Autocomplete.vue
@@ -633,7 +633,7 @@ export default {
   font-family: sans-serif;
   position: absolute;
   list-style: none;
-  background: $color-gray--lighter;
+  background: $color-white;
   padding: 0;
   margin: 0;
   display: inline-block;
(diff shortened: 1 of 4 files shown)


edwh:
Thanks.  This one sounds like I need to think about it as I review it.

JayyajGH:
Yeah agreed.  I almost created it as a draft to prompt that.  It does feel slightly messy but I think feels better than it did before.

PR #320, opened 2020-10-12
Feature/add word highlight to search

JayyajGH:
Take a look and see what you think of the search highlighting.  

The autocomplete code is a bit of a mess - I'm guessing because it was ported from another repo.  Do we ever use the "(data, label)" bit as opposed to the "(data,anchor)" - for example in the ul dropdown list part?  There is a lot of styling attached to that and also quite a bit of styling that says "backwards compatibility".  I'm a bit scared to just remove it!

### DIFF
--- a/components/Autocomplete.vue
+++ b/components/Autocomplete.vue
@@ -48,7 +48,13 @@
             <!-- eslint-disable-next-line -->
             <div v-if="onShouldRenderChild" v-html="onShouldRenderChild(data)" />
             <div v-if="!onShouldRenderChild">
-              <b class="autocomplete-anchor-text">{{ deepValue(data, anchor) }}</b>
+              <Highlighter
+                :text-to-highlight="deepValue(data, anchor)"
+                :search-words="[type]"
+                highlight-class-name="highlight"
+                auto-escape
+              />
+              <!--<span class="autocomplete-anchor-text">{{ deepValue(data, anchor) }}</span>-->
               <span class="autocomplete-anchor-label">{{ deepValue(data, label) }}</span>
             </div>
           </a>
@@ -74,8 +80,12 @@
 /* eslint-disable */
 
 import cloneDeep from 'lodash.clonedeep'
+const Highlighter = () => import('vue-highlight-words')
 
 export default {
+  components: {
+    Highlighter
+  },
   props: {
     id: String,
     name: String,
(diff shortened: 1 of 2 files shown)


JayyajGH:
I'm now thinking about the other requirements i.e. different text colours and sizes. Making this work alongside the Vue highlighter is likely to not be simple.  It's that component that splits out the text into separate spans.

So "Bristol Airport, Hyatts Wood Road, Flax Burton" would be split by the highlighter into:
1: "Bristol"
2: "Airport, Hyatts Wood Road, Flax Burton"

I can't easily think of a way of doing this without writing our own highlighter.  I'll take a look at the highlighter (and maybe some other highlighters) to see if they provide a way of marking text in this way.

edwh:
The highlighting looks ok in this.  Do you want me to release this as it stands?

Autocomplete is a mess, indeed.  I picked it up from somewhere else (credit in the code) and then started changing it when we needed slightly different function.  I'm in favour of aggressively simplifying this component down to what we need - since we don't use it in many places it's easy to test.   We only use it in a few places, and we always pass label="".

There's also AutocompleteLocal, which is simpler but shouldn't really be separate.

JayyajGH:
The reason I left it as a draft was because it didn't really complete the requirements.

Originally you were asking for parts of the text to be black and then stuff after the comma etc. to be in a smaller font etc.  That's the stuff that is going to be quite tricky if we are using the current text highlighter.  I've had a bit of a look around and I haven't found anything.  The only way I can think of achieving this (apart from writing our own) is to do it in 2 passes.  Run something over the text that splits it out in pre/post comma and then use run the highlighter over each section.  It wouldn't be the prettiest of things so I guess it depends on how much you want that style.  I personally quite like it as it is.  Looking at the google maps one I actually had to check in the code to see if the font was any smaller so I think it's quite subtle.

I was concerned that I might have broken something with the label part of the autocomplete as I'd removed one of the styles for it (labelled backwards compatible) and hadn't really tested anything around it.  Sounds like it's safe then.

It's probably easier to pull the autocomplete refactoring out of this piece of work I think.  I'll add it to the list!

edwh:
I think with the bold highlighting the requirement to have some of it in 
smaller font is weaker.

So I think I can release this one as it stands - incrementally better.  
Right?

JayyajGH:
I like the way it looks now so I'm good to go. I'll un-draft it.

PR #323, opened 2020-10-19
Feature/accessibility updates to selects

JayyajGH:
Add sr-only labels to the two select fields.  I've changed the GroupSelect component to have a label field but it's not mandatory (for now).  It felt right to have this in the same component alongside the select field itself.  i.e. the select field in GroupSelect should always have a label so it should be contained together.

### DIFF
--- a/components/AdaptiveMap.vue
+++ b/components/AdaptiveMap.vue
@@ -85,9 +85,11 @@
           Search Filters
         </h2>
         <div v-if="filters" variant="info" class="p-2 border border-info bg-white filters">
-          <GroupSelect v-model="selectedGroup" all :all-my="false" />
+          <label for="communitiesList" class="sr-only">Communities to view</label>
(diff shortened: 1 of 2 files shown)


edwh:
We use GroupSelect all over the place.  Don't you need to default labelSROnly to true to handle that?

JayyajGH:
I've done it so that nothing gets added if you don't pass in a label.  If you pass in a label with labelsronly missing or false then the default is the label gets displayed which I thought would be the norm.  The labelSROnly flag was just to state whether or not the label you are passing it should be hidden or not.

All the existing uses of GroupSelect won't pass either field and so the label won't get rendered at all.

Or am I missing something?

edwh:
No, I'm being daft.  Looks fine, then.

JayyajGH:
Ha!  Always worth raising in case I've totally forgotten something.  I tested the various options and it looked ok but always nice to have a second opinion.

JayyajGH:
Ideally we should now do this in loads of other places but that's another piece of work.

PR #326, opened 2020-10-23
Allow the image to be accessed via the keyboard

JayyajGH:
Something I noticed when looking into the freegled stuff.  You couldn't expand the image using the keyboard as it didn't have a tabstop or any keyboard event.

### DIFF
--- a/components/Message.vue
+++ b/components/Message.vue
@@ -37,7 +37,7 @@
             See details and reply <v-icon name="angle-double-right" />
           </b-button>
         </div>
-        <div v-if="attachments && attachments.length > 0" class="clickme card-header-image__wrapper" tabindex="0" @click="showPhotos" @keyup.enter="showPhotos">
+        <button v-if="attachments && attachments.length > 0" class="card-header-image__wrapper p-0 border-0" @click="showPhotos">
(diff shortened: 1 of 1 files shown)


JayyajGH:
Actually hold off on this one for a bit.  I think maybe making it a button with an included image would be a better option and then we're not fighting HTML.

JayyajGH:
OK I've now made this a button.  There wasn't some crazy reason why this needed to be a div was there?

This change is all part of showing the freegled items.  I'm currently trying to find the best way to stop clicks/keyboard interaction on a freegled item and the image was causing issues.  Fundamentally it's because a button wouldn't smell any sweeter if it was a div.  In fact you would need to add an artificial smell to it.

edwh:
I don't think there was any reason for it not to be a button.  And as 
you point out, any such reason would be crazy.

JayyajGH:
Some might also say that removing auto-capitalisation from a control to make it work on Safari would be crazy too...

JayyajGH:
So this one is now good to go.  Definitely worth you taking a look though to check I haven't missed anything.  It should now have tabbing and keyboard control out of the box.

edwh:
Checked it out, had a play, looks good.

PR #328, opened 2020-10-27
Feature/chat html tidy

JayyajGH:
OK this is the refactor of the chat message box to remove the duplicated HTML and to make it accessible by adding a label.  Do you want to take a look and see what you think?  I've so far only tested it on Chrome on W10.  I'll give it a test tomorrow on other devices.

I currently can't test the alt-shift combination as on Windows 10 it appears to be the key combination to change language.  There wasn't an obvious way of switching this off but I'll investigate further.

### DIFF
--- a/components/ChatFooter.vue
+++ b/components/ChatFooter.vue
@@ -51,7 +51,7 @@
           placeholder="Type here..."
           rows="3"
           max-rows="8"
-          :autocapitalize="enterNewLine ? 'none' : false"
+          :autocapitalize="enterNewLine ? false : 'none'"
           @focus="markRead"
           @[keydownEventIfNotEnterForNewLine].enter.prevent=""
           @[keydownEventIfNotEnterForNewLine].enter.shift.exact.prevent="newline"
--- a/components/ChatFooter.vue
+++ b/components/ChatFooter.vue
@@ -224,7 +224,7 @@ export default {
     keydownEventIfNotEnterForNewLine() {
       return this.enterNewLine ? null : 'keydown'
     },
-    keyupeventIfNotEnterForNewLine() {
+    keyupEventIfNotEnterForNewLine() {
       return this.enterNewLine ? null : 'keyup'
     }
   }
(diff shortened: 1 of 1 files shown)


JayyajGH:
I've now tested this on the following:
Chrome on W10
IE11 on W10
Chrome on Galaxy S9
Chrome on Galaxy S20
Safari 12 on iPhone 8.
Safari 13.1 on Catalina

Is there anywhere else that would be useful to test?  I'm struggling to get many of the iPhones working at all.

edwh:
Sounds ok, but what's happening with the iPhones?

JayyajGH:
I'm getting the website loading but just doing the spinny loading/stuck here thing which never goes away and the sign in button never appears.

edwh:
Console error?

JayyajGH:
Nothing.... A few of the ones I managed to test did it to but after refreshing about 30 times it did load in the end.  Others I couldn't get to load.

edwh:
I'm scared of the way you've done it :-).

You've got a handler attached for a null event as a way of avoiding having two DOM elements.  That's the kind of unusual thing that might break on some platforms, he said nervously.  Then again, https://vuejs.org/v2/guide/syntax.html#Dynamic-Arguments seems to suggest that null is an accepted way to remove a binding.

I think the code would be easier to read, and less weird, if you had two div wrappers with a label in each.  Am I missing something?

JayyajGH:
Yeah I did that because of that exact link that you saw too!  This is Vue logic I suppose rather than anything to do with Javascript or the browser so devices/browsers shouldn't matter.  Vue should just not attach a handler to the event in the case it returns null.

The issue with two divs is you then need to duplicate the code.  If you update the classes or logic for one then you also need to update for the other etc.  I guess I just really hate the duplication.

Although I can see where you're coming from.  If you're really not comfortable with it then happy to update it.  I've learnt about dynamic arguments so I've gained from it. :-)

edwh:
I'd prefer it that way.  It feels a bit like code that's too pleased 
with itself for its own good :-).  I prefer a bit of duplication 
sometimes if the alternative is that people have to scratch their heads 
to work out why it is how it is.

...though now I know about the dynamic thing too, which is interesting.

JayyajGH:
Yeah no worries.  I'll park it for now and just add some labels.  In the office tomorrow so it's likely to be monday now

PR #330, opened 2020-11-04
Refactor the notification options into a separate component.

JayyajGH:
This is the next phase of tidying up default.vue

I've currently set it to draft as I haven't fully tested it yet but thought I'd get your opinion on it.

At the moment I've just taken out anything to do with notifications and put it in the component.  It's still got two separate HTML sections; one for large and one for small screens.

One part I'm unsure about is I've left the unreadNotificationCount property in default.vue.  Because it needs to be real-time, it doesn't really work living in the component and raising an event to the parent as there is nothing currently to trigger it.  I can't think of another way to handle it which isn't horrible.  Thinking about it now I'm not actually sure if the head method is reactive anyway.  I'll check on that...

### DIFF
--- a/components/NotificationOptions.vue
+++ b/components/NotificationOptions.vue
@@ -1,82 +1,43 @@
 <template>
-  <div>
-    <!-- Notification menu option for large screens -->
-    <b-nav-item-dropdown
-      v-if="!simple"
-      class="white text-center notification-list d-none d-xl-block"
-      toggle-class="notification-list__dropdown-toggle"
-      menu-class="notification-list__dropdown-menu"
-      lazy
-      right
-      @shown="loadLatestNotifications"
-    >
-      <template slot="button-content">
-        <div class="position-relative text-center small">
-          <v-icon name="bell" scale="2" class="notification-list__icon" />
-          <b-badge v-if="unreadNotificationCount" variant="danger" class="notification-badge">
-            {{ unreadNotificationCount }}
-          </b-badge><br>
-          <span class="nav-item__text">Notifications</span>
-        </div>
-      </template>
-      <b-dropdown-item class="text-right" link-class="notification-list__item">
-        <b-btn variant="white" size="sm" @click="markAllRead">
-          Mark all read
-        </b-btn>
-      </b-dropdown-item>
-      <b-dropdown-divider />
-      <b-dropdown-item v-for="notification in notifications" :key="'notification-' + notification.id" class="p-0 notpad" link-class="notification-list__item">
-        <Notification :notification="notification" class="p-0" @showModal="showAboutMe" />
-      </b-dropdown-item>
-      <infinite-loading :distance="distance" @infinite="loadMoreNotifications">
-        <span slot="no-results" />
-        <span slot="no-more" />
-        <span slot="spinner">
-          <b-img-lazy src="~/static/loader.gif" alt="Loading" />
-        </span>
-      </infinite-loading>
-    </b-nav-item-dropdown>
-
-    <!-- Notification menu option for small screens -->
-    <b-dropdown
-      v-if="loggedIn"
-      class="white text-center notification-list mr-2 d-block d-xl-none"
-      variant="transparent"
-      toggle-class="notification-list__dropdown-toggle"
-      menu-class="notification-list__dropdown-menu"
-      lazy
-      right
-      @shown="loadLatestNotifications"
-    >
-      <template slot="button-content">
-        <div class="position-relative">
-          <v-icon name="bell" scale="2" class="notification-list__icon" />
-          <b-badge v-if="unreadNotificationCount" variant="danger" class="notification-badge">
-            {{ unreadNotificationCount }}
-          </b-badge>
-        </div>
-      </template>
-      <b-dropdown-item class="text-right">
-        <b-btn variant="white" size="sm" @click="markAllRead">
-          Mark all read
-        </b-btn>
-      </b-dropdown-item>
-      <b-dropdown-divider />
-      <b-dropdown-item v-for="notification in notifications" :key="'notification-' + notification.id" class="p-0 notpad">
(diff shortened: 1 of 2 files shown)


edwh:
Moving this out into another component is clearly a good idea.

Passing the parent refs seems like a right old hack.  We only use aboutMeModal - couldn't that just move into here too?

I'm fairly sure head is reactive - I see my tab titles change.  One way to handle this would be to add a watch() inside your new component, and then use that to trigger an update:X event up to the parent?

JayyajGH:
I still think this needs future refactoring as it's still not great but once default.vue is smaller it'll be easier to see the wood for the trees.

Yeah good point re aboutmeModal.  I'll move it in.  It's not the prettiest of bits of code but does seem to be mentioned quite a lot in the wild.

But what would you watch on?  There isn't any specific property that is up to date.  I suppose you could watch the store property.  I've not done that before but it's reactive so I can't see why not.  I'll give that a go.

JayyajGH:
Doh.  I can just watch the existing computed property....  I think I was overcomplicating this in my head

JayyajGH:
Written some code but I don't have much time to test today to I'll continue tomorrow.  I need to post something in live chitchat to test don't I?

edwh:
Yes, that's a good way.  Have two users and have a little chat with 
yourself, then delete the thread fairly rapidly so it doesn't get mailed 
out to everyone.

JayyajGH:
OK so I've now given this a test on both desktop and mobile and it looks fine.  Would be good to refactor this more as there is a lot of duplication in here but that will be next.

My main concern was with the updating in the parent component of the notification count.  This coding worked first time which always makes me nervous!  I've not really done much else.  

Would be good for you to give it a proper test though before it moves anywhere.  I'll move it to ready for review...

JayyajGH:
@edwh - Also, where do you do the mobile/desktop test that you can use from JS?  With the dummy element and check whether the element was hidden/shown?  I wanted to steal it rather than writing something different.

JayyajGH:
Actually can you hold off on this one.  Although it works, I'm not happy with the semantics.  Github doesn't seem to have an option to make something not ready for review.

edwh:
Okey-dokey.

edwh:
Sorry, I missed the question about JS.

I think this is the thing you're looking for:


edwh:
Every now and again, Ctrl+V sends my mails.

<div id="sizer" ref="sizer" class="d-none d-lg-block" />

From: "Edward Hibbert" <edward@ehibbert.org.uk>
To: "Freegle/iznik-nuxt" 
<reply@reply.github.com>
Sent: 09/11/2020 18:39:02
Subject: Re[2]: [Freegle/iznik-nuxt] Refactor the notification options 
into a separate component. (#330)

JayyajGH:
Ah I deleted the question as I found the code in message.vue.  But that was slightly different from this...    It works fine but didn't work as you dragged the screen around.  Only on page load.
 
This code only watches for a route change though doesn't it?  It's not dynamic as you change the screen size?  Fundamentally the same as message.vue with a route change watcher.  I guess there isn't anything reactive to watch on.   It would have to be some kind of resize observer.

So to recap:
1. I can pull the existing notification code out into a component but due to the semi-duplicated small/large elements then there needs to be a wrapper div.  This puts a div into the ul rather than an li.
2. I then changed the code to only have one element but using a dynamic component and moving the media query logic into the parent.  This resolves the semantic issues with the ul element.
3. The remaining issue is as you drag the screen around then the change doesn't trigger anything so things don't appear as they should.

Am I missing something with your existing code?  Does it work in the way that we require?

edwh:
Yes, it's probably not dynamic. But with a timer...?

I'll have a look at the code tomorrow.

edwh:
I'm losing track slightly.  But could you have two copies of the 
component in the parent, with an option saying whether it's small/large, 
and then display/hide the copies in the parent using the standard 
classes?

JayyajGH:
Yes that's pretty much what I've got now but currently the "sizer" method you've used elsewhere doesn't rejig when you change the screen size.  I'll try out a timer and an observer to see how they feel.

edwh:
Might want to move the sizer into a separate component.  Then it can 
create/destroy its own timer, and emit events.

JayyajGH:
Yeah having something reusable would be useful

edwh:
You could make it into a separate npm package, for fun.

JayyajGH:
I'm still not that sold on the timer idea.  My current line of thinking is can we use something like the bootstrap v-b-visible directive or something similar.  I'll have a poke around with some stuff and report back...

JayyajGH:
In a way it seems like we are fighting bootstrap here.  If I wasn't using bootstrap then I would set the breakpoint values in CSS custom properties and then they would automatically be available in JavaScript.  It almost seems crazy to be talking about timers and intersection observers just to get a breakpoint value.

One of the things on the to-do list is to sort out the bootstrap inheritance chain.  That would basically involve tweaking the way nuxt imports it's scss.  What we could do is store the breakpoints in CSS custom properties and then pass them into bootstrap.  Then we get the best of both worlds and we can access them in javascript.

JayyajGH:
https://bootstrap-vue.org/docs#using-custom-bootstrap-scss

edwh:
If all we want is to get the properties into JS then we could do that as 
you suggest (would sounds better) but even just by doing variations on 
the sizer trick and putting them in the store.

But isn't the issue that we want to spot when the current breakpoint 
changes?

I could be missing something.

JayyajGH:
The actual goal (lost somewhere in the ever increasing thread) is to display component A or component B based on screensize. For small screens display component A and for large screens display component B.  Usually you would do this with a regular CSS media query but I'm using a dynamic component which is driven by JavaScript.

So I have some HTML:
<component v-is="componentType">

and a computed property
componentType() {
return screen size >800px ? 'componentA' : 'componentB'
}

edwh:
Let's switch to Slack.

PR #331, opened 2020-11-05
Add an accessible label to the chat messages

JayyajGH:


edwh:
This looks fine and I'll merge it.

One thing that always makes me uneasy is whether the addition of an sr-only label will mess up layout.  For example, if we were using grid, the introduction of an extra DOM element could quite easily break things.   This makes me wonder whether we should have a convention of wrapping the label and original element inside a div, so that the DOM structure is less changed.  Maybe not, though, maybe just watch out for grid.

JayyajGH:
Yeah if you were using auto placement then it would leave a slot for it I think and therefore mess up the grid.  Not sure what it would do if you explicitly placed items.  I think it would probably just ignore it but I'm not sure without checking.

A counter argument would be that all fields should have visible labels.  A lot of people consider using placeholder text a bad idea without a label. Also adding an extra div could just be unnecessary most of the time. so just polluting the HTML.

My personal view would be that it's probably something that should just be considered in each instance but I don't have particularly strong views on it.

edwh:
We'll do nothing for now.

I very much take the point about always having labels, though I often 
struggle for screen space to do that on mobile.  I'm trying to get away 
from just using placeholders.

PR #333, opened 2020-11-11
Add some space between the grid columns

JayyajGH:
Add some breathing space to the gird items. 

grid-column-gap and grid-gap have actually been deprecated but we still need to support the older browsers.


edwh:
Ah, useful.  I have been faffing with putting explicit columns in as spacers in some place as I hadn't noticed grid-column-gap.

PR #334, opened 2020-11-16
Refactor the notification options into a component

JayyajGH:
This is a rethink of the notification options component.

Generally I'm happy with the component now although I need to give it some more testing hence the draft PR.  The only point of contention is the about me modal which I've been struggling with a bit.

So at the moment the component can only have a single element due to the way the parent is set up.  The parent in some cases is a UL and so the component can't have a wrapper DIV or it breaks the structure.

Thinking about ownership, it doesn't seem right for the about me modal to live in the notification component.  It has got nothing to do with notifications.  It just happens to be called from that component.  This modal is called from other places and these places currently also define a version of the modal. (Potential future refactor...)

This PR leaves the about me modal in the default layout and just raises an event from the notifications component to show it.  To me this feels much nicer as it keeps the notification component clean and focused on one thing.

Thoughts?

### DIFF
--- a/components/ChatMenu.vue
+++ b/components/ChatMenu.vue
@@ -31,6 +31,7 @@ export default {
   },
   computed: {
     chatType() {
+      // A different component needs to be created depending on the context in which it's used
       return this.smallScreen ? 'a' : 'b-nav-item'
     },
     chatCount() {
--- a/components/ChatMenu.vue
+++ b/components/ChatMenu.vue
@@ -41,7 +41,7 @@ export default {
   },
   watch: {
     chatCount: function() {
-      this.$emit('chat-count', this.chatCount)
+      this.$emit('update:chatCount', this.chatCount)
     }
   },
   methods: {
--- a/components/ChatMenu.vue
+++ b/components/ChatMenu.vue
@@ -8,8 +8,10 @@
     @click="toChats"
   >
     <div class="position-relative">
-      <v-icon name="comments" scale="2" /><br>
-      <span v-if="!smallScreen" class="nav-item__text">Chats</span>
+      <v-icon name="comments" scale="2" />
+      <div v-if="!smallScreen" class="nav-item__text">
+        Chats
+      </div>
       <b-badge v-if="chatCount" variant="danger" class="chatbadge">
         {{ chatCount }}
       </b-badge>
@@ -65,5 +67,10 @@ export default {
   position: absolute;
   top: 0px;
   left: 25px;
+}
+
+svg.fa-icon {
+  height: 32px;
+  margin: 0;
 }
 </style>
--- a/components/ChatMenu.vue
+++ b/components/ChatMenu.vue
@@ -1,0 +1,69 @@
+<template>
+  <component
+    :is="chatType"
+    :id="smallScreen ? 'menu-option-chat-sm' : 'menu-option-chat'"
+    :class="smallScreen ? 'text-white mr-3 position-relative' : 'text-center small p-0'"
+    href="#"
+    aria-label="chats"
+    @click="toChats"
+  >
+    <div class="position-relative">
+      <v-icon name="comments" scale="2" /><br>
+      <span v-if="!smallScreen" class="nav-item__text">Chats</span>
+      <b-badge v-if="chatCount" variant="danger" class="chatbadge">
+        {{ chatCount }}
+      </b-badge>
+    </div>
+  </component>
+</template>
+
+<script>
+export default {
+  name: 'ChatMenu',
+  props: {
+    smallScreen: {
+      type: Boolean,
+      required: false,
+      default: false
+    }
+  },
+  computed: {
+    chatType() {
+      return this.smallScreen ? 'a' : 'b-nav-item'
+    },
+    chatCount() {
+      // Don't show so many that the layout breaks.
+      return Math.min(99, this.$store.getters['chats/unseenCount'])
+    }
+  },
+  watch: {
+    chatCount: function() {
+      this.$emit('chat-count', this.chatCount)
+    }
+  },
+  methods: {
+    toChats(e) {
+      if (e) {
+        e.preventDefault()
+        e.stopPropagation()
+        e.stopImmediatePropagation()
+      }
+
+      // Ensure we have no chat selected.  On mobile this will force us to show the chat list.
+      this.$store.dispatch('chats/currentChat', {
+        chatid: null
+      })
+
+      this.$router.push('/chats')
+    }
+  }
+}
+</script>
+
+<style scoped lang="scss">
+.chatbadge {
+  position: absolute;
+  top: 0px;
+  left: 25px;
+}
+</style>
--- a/components/ChatMenu.vue
+++ b/components/ChatMenu.vue
@@ -8,7 +8,7 @@
     @click="toChats"
   >
     <div class="position-relative">
-      <v-icon name="comments" scale="2" />
+      <v-icon name="comments" scale="2" class="chat__icon" />
       <div v-if="!smallScreen" class="nav-item__text">
         Chats
       </div>
@@ -69,7 +69,7 @@ export default {
   left: 25px;
 }
 
-svg.fa-icon {
+.chat__icon {
   height: 32px;
   margin: 0;
 }
--- a/components/NotificationOptions.vue
+++ b/components/NotificationOptions.vue
@@ -100,8 +100,6 @@ export default {
       if (this.complete) {
         $state.complete()
       } else {
-        this.busy = true
-
         try {
           const currentCount = this.notifications.length
 
@@ -117,10 +115,8 @@ export default {
           } else {
             $state.loaded()
           }
-          this.busy = false
         } catch (e) {
           console.error(e)
-          this.busy = false
           $state.complete()
         }
       }
--- a/components/NotificationOptions.vue
+++ b/components/NotificationOptions.vue
@@ -45,6 +45,7 @@ const Notification = () => import('~/components/Notification')
 const InfiniteLoading = () => import('vue-infinite-loading')
 
 export default {
+  name: 'NotificationOptions',
   components: {
     InfiniteLoading,
     Notification
--- a/components/NotificationOptions.vue
+++ b/components/NotificationOptions.vue
@@ -1,0 +1,182 @@
+<template>
+  <component
+    :is="notificationType"
+    class="white text-center notification-list"
+    :class="{'mr-2' : smallScreen}"
+    :variant="smallScreen ? 'transparent' : ''"
+    toggle-class="notification-list__dropdown-toggle"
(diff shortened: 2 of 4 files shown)


edwh:
That sounds fine.

In general we could have a bunch of modals which get triggered from 
different places.  The microvolunteering stuff I'm working on is a bit 
like that.  So if you wanted to feel better about it, you could create a 
wrapper component for that and listen on $root as an event bus.

JayyajGH:
I'll leave this as it is for now and look into the modals as a phase 2 I think.  I'll give this a proper test and then un-draft it.

JayyajGH:
This is now ready for review.  I have tried to leave most of the code as it was e.g. the existing JS and CSS. There is some more refactoring in those areas but that can be done as another piece of work.

edwh:
Thanks - I'll have a look today.

JayyajGH:
No rush.  It's part of a long journey to tidy default.vue...

edwh:
Can I dump something into that?

Someone pointed out that on pages with the sidebars, you get the 
"About/Terms" etc links.  But on pages where we don't have these (for 
clutter reasons) you don't.  That's probably wrong.  Those links are 
really site wide so they should be in the layout, I think.

JayyajGH:
So every page should have the about/terms displayed on the sidebar?  We could just add them in to the missing pages as a quick fix or just wait until the layout pages are refactored.

edwh:
Yes.

It's not urgent, so I'd rather it was part of the layout, assuming you 
agree that's where it should be.

JayyajGH:
0k cool.  Do they need to be so obvious?  Just looking at other infinite scroll websites (FB and twitter) and they seem to hide their terms etc. away in menus.  Is that an option?  Especially as we already hide them on the simple view

edwh:
They're hidden on smaller screens, too, of course.

We have links in the Help section, which you'd have thought would be 
enough.  But we do get a steady stream of people who are asking for and 
cannot find them.  For example we recently had someone who wanted to 
donate, couldn't find the donate link, googled "freegle donate", and 
then contacted us to say it was hard to find.

So in the cases where the sidebar is visible but empty, I'd prefer it if 
they were there.

JayyajGH:
In that case do we need all of them?  Can we just have donate on the main screen and hide the others?

edwh:
We get questions about the others too.

JayyajGH:
Thought I'd check! :-)

JayyajGH:
I've now got the chat stuff pulled out into a component as well.  Do you want to deal with that separately or with this?

edwh:
With this - not got to it yet.

JayyajGH:
Was there any reason you can remember for adding a "to" attribute on the chat anchor?  It's all run off JS and there is preventdefault set so I can't really see a use for it.  I've just added a href="#" to both small and large screens. (Although technically it isn't required but bootstrap adds it to b-nav-item so I've added it for consistency)

edwh:
Where precisely?

JayyajGH:
In default.vue there is a b-nav-item for chats with an id of "menu-option-chat"

edwh:
I think that's a hangover from earlier code.

JayyajGH:
OK cool.  It works ok without it. :-)

Although just noticed a bit of an accessibility issue with the mobile menu so I'll fix that whilst I'm at it

JayyajGH:
I've just pushed the chats refactor too.  I haven't given this a great test yet though.  It's had minimal testing.

edwh:
Ok.

It's good that you're doing this kind of stuff.  It would never reach 
the top of my list, but it stops code decay.

JayyajGH:
It's like how the Vue team used to work.  The core team (you and Chris) work on new features and the community (me) work on bugs and smaller improvements. :-)

components/NotificationOptions.vue
```diff
@@ -0,0 +1,185 @@
...
+    },
+    smallScreen: {
+      type: Boolean,
+      required: false,
+      default: false
+    }
+  },
+  computed: {
+    notificationType() {
+      return this.smallScreen ? 'b-dropdown' : 'b-nav-item-dropdown'
```
edwh, line 67: Can you comment saying why we do this?

components/ChatMenu.vue
```diff
@@ -0,0 +1,76 @@
...
+      return this.smallScreen ? 'a' : 'b-nav-item'
+    },
+    chatCount() {
+      // Don't show so many that the layout breaks.
+      return Math.min(99, this.$store.getters['chats/unseenCount'])
+    }
+  },
+  watch: {
+    chatCount: function() {
+      this.$emit('chat-count', this.chatCount)
```
edwh, line 43: Wouldn't it be better to emit update:chat-count so that the parent can use .sync?

components/NotificationOptions.vue
```diff
@@ -0,0 +1,185 @@
...
+        'notifications/getCurrentListInDescendingDateOrder'
+      ]
+    },
+    unreadNotificationCount() {
+      return this.$store.getters['notifications/getUnreadCount']
+    }
+  },
+  watch: {
+    unreadNotificationCount: function() {
+      this.$emit('unread-notification-count', this.unreadNotificationCount)
```
edwh, line 80: Emit update:?

components/NotificationOptions.vue
```diff
@@ -0,0 +1,185 @@
...
+    loadMoreNotifications: function($state) {
+      const currentCount = this.notifications.length
+
+      if (this.complete) {
+        $state.complete()
+      } else {
+        this.busy = true
+        this.$store
+          .dispatch('notifications/fetchNextListChunk')
+          .then(() => {
```
edwh, line 107: When I wrote this I was clueless about await.  While we're restructuring this code, could we rework it to use that?

layouts/default.vue
```diff
@@ -392,9 +307,8 @@ export default {
       console.log('Start NCHAN from mount')
       this.startNCHAN(me.id)
 
-      // Get notifications and chats and poll regularly for new ones.  Would be nice if this was event driven instead but requires server work.
+      // Get chats and poll regularly for new ones.  Would be nice if this was event driven instead but requires server work.
```
edwh, line 310: This comment suggests a timer which has now moved into the store, in notifications.js.  Can you move the comment there?

components/ChatMenu.vue
```diff
@@ -0,0 +1,76 @@
...
+  props: {
+    smallScreen: {
+      type: Boolean,
+      required: false,
+      default: false
+    }
+  },
+  computed: {
+    chatType() {
+      return this.smallScreen ? 'a' : 'b-nav-item'
```
edwh, line 34: Can you comment why we do this?

layouts/default.vue
```diff
@@ -392,9 +307,8 @@ export default {
...
       this.fetchLatestChats()
```
edwh, line 311: We've lost a call to updateNotifications here. Is that intentional?

edwh (commented):
This rips a good chunk of code out, which is nice.  A few comments.

components/NotificationOptions.vue
JayyajGH: Do you mean technically as in how the code is used or why we have two different styles of menu?

components/ChatMenu.vue
JayyajGH: I avoided using sync as it's been deprecated in v3 hasn't it?

components/NotificationOptions.vue
JayyajGH: Would something like this make sense?

// A different component needs to be created depending on the context in which it's used

layouts/default.vue
JayyajGH: Which one?  The updateNotifications call which was in mounted in default.vue has been moved into the mounted of notificationoptions.vue.

Although I have noticed that I can also move the fetchchats in to the relevant component....

edwh:
.sync itself is going, but I think you still emit the update event.  
That's what 
https://v3.vuejs.org/guide/migration/v-model.html#using-v-bind-sync 
suggests to me.

JayyajGH:
Ah so it's actually quite similar.  I obviously hadn't read it properly!

edwh:
In both cases, why we're using different components.  It's just not 
locally obvious from looking at that component, and because it's unusual 
to do this I think it bears a bit more commenting.

JayyajGH:
I'm actually still a bit confused as to why we are using different component!  I was going to start looking at that once I could see the wood for the trees in default.vue.  In chats for example, small screens use b-nav-item which creates an anchor and large screens directly use an anchor.

JayyajGH:
Looks like we could pull fetchLatestChats out into the store in the same way as notifications works.

components/NotificationOptions.vue
JayyajGH: Just checked in some code to use await.  It looks much tidier.  You probably should take a proper look at this bit.  I think I've got everything but was unsure about some of the infinite-loading values.  e.g. Do we still need the busy property?

edwh:
Catching up:

- I can't remember why we use different components either, which is what makes me think it either needs a comment or might be wrong and then could be removed.
- For chats, I think there is more refactoring that is needed, but it's not clear what.  Maybe move chat updating out of pages into the store, as you say, but there will be subtleties about when we need to ask for updates.  Definitely not part of this - you're tidying up the default vue here.

So this might be nearly there?

JayyajGH:
For now I've added the comment:

// A different component needs to be created depending on the context in which it's used

Does that sort of cover it for now?  Once default.vue is stripped down to it's basics then we can start looking into how the menu itself works and maybe working out why it is how it is.  Moving it out into a component isn't making it any worse than it was before or any more confusing.  If anything it's actually making it more obvious.

JayyajGH:
Chats - My new component is basically for the menu and that's it.  I think the next step will be to move the chat update stuff out into the store maybe but I can look at that as a separate piece of work.

JayyajGH:
You did mention earlier "We've lost a call to updateNotifications here. Is that intentional?"  I think they've all been included.  Could you double check that please.  I'm hoping you're wrong! :-)

JayyajGH:
And also could you take a closer look at the promise/await refactor I did please.  Mainly to check that I got the vue-infinite stuff correct rather than the promises.

edwh:
I've merged master to make it easier to compare.  If I do a compare against master I see this:

image

JayyajGH:
Ah yes.  I've moved that into the mounted method in the component.

edwh:
So you have.

The await stuff looks fine.  The busy field is not needed if it ever was, so you could lose that.  Then I think it's good to go.

JayyajGH:
In that case we're good to go!

edwh:
Phew!

JayyajGH:
:-)  That was quite painful.  At least we got there in the end.  default.vue is looking much more manageable already

PR #336, opened 2020-11-19
Move the regular chat fetching into the store

JayyajGH:
Move the timer based chat fetches into the store.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -217,7 +217,6 @@ export default {
     return {
       complete: false,
       distance: 1000,
-      chatPoll: null,
       nchan: null,
       logo: require(`@/static/icon.png`),
       timeTimer: null,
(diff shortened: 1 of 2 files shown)


edwh (commented):
I think once you move this into the store it will start to run for ModTools as well, and therefore you'll need to make the equivalent changes to the modtools layout?

JayyajGH:
I'm made the change to MT already but not created a PR for it. When I next switch the PC on then I'll add a PR and they can go together.

JayyajGH:
I'll delete this PR and include these changes in the MT branch as that merge will be easier.

JayyajGH:
@edwh - Thinking about this I don't think this would have automatically run for MT as it's an action in the store which would need dispatch to set it off.  The timer is created within the action.  Regardless I have now merged my two changes so I'll create a PR for the combined change.

edwh:
I think you're right.

PR #337, opened 2020-11-23
Chat menu refactoring for default and modtools layouts

JayyajGH:
- Tweaks to the chatmenu component so it's more generic
- Move chat fetching into the store rather than the layout components
- Use the chatmenu component from MT

### DIFF
--- a/components/ChatMenu.vue
+++ b/components/ChatMenu.vue
@@ -1,15 +1,14 @@
 <template>
   <component
     :is="chatType"
-    :id="smallScreen ? 'menu-option-chat-sm' : 'menu-option-chat'"
-    :class="smallScreen ? 'text-white mr-3 position-relative' : 'text-center small p-0'"
+    :class="{'text-white' : !isListItem}"
     href="#"
     aria-label="chats"
     @click="toChats"
   >
-    <div class="position-relative">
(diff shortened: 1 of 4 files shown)


edwh:
On ModTools, the Chat option is always greyed out.

This is not your bug, but in ModTools the chat fetch ought to also include the little-used Mod2Mod chats.  Can you add that in?

Code looks ok.

JayyajGH:
That's odd.  It wasn't like that before so I must have subsequently broken it.  I'll take a look.

Yup I'll add the other change in too.

JayyajGH:
For the Mod2Mod change do you mean for everyone? 

Currently the chat types are: chattypes: ['User2User', 'User2Mod']

so do you mean just add Mod2Mod to that list?

edwh:
No, that should only be added in the MT case.  So you may need to use 
rootGetters to access the modtools flag.

JayyajGH:
Ah ok  will do.

JayyajGH:
OK I've now fixed the greyed out chat icon and also added in the extra chat type for mods.  The greyed out icon is a bit messy really.  I'm hoping in the near future it will be possible to have just one <nav-link> element here and not a choice of nav-link or anchor.

PR #338, opened 2020-11-25
Refactor the nchan logic into a mixin

JayyajGH:
Here it is... initially just for discussion

My main worries are around "me".  The "me" used in mounted needs to be re-defined as it won't pick up the "me" from the global mixin.  The other oddity is there is a watch over "me" as well.  This appears to work using the one from the me.js mixin but I'm a bit uncomfortable watching something that you are just assuming exists elsewhere.  Therefore I have added a local "me" property.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -194,10 +194,10 @@ import LocalStorageMonitor from '~/components/LocalStorageMonitor'
 import BouncingEmail from '~/components/BouncingEmail'
 import NotificationOptions from '~/components/NotificationOptions'
 import ChatMenu from '~/components/ChatMenu'
+import nchanHelper from '@/mixins/nchanHelper'
 
 const AboutMeModal = () => import('~/components/AboutMeModal')
 const ChatPopups = () => import('~/components/ChatPopups')
-const NchanSubscriber = require('nchan')
 const ExternalLink = () => import('~/components/ExternalLink')
 
 export default {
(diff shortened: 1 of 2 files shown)


edwh:
If you're worried about that, why not use a different name for me which won't clash with the global one?

JayyajGH:
It wasn't the extra "me" that concerned me.  I just didn't like having the watch over "me" when there was no "me" defined in that component.  That is why I added a local one.  Does that make sense?

I think I prefer it with the same name.  Although I'm not sure which one takes precedence.  I know the components methods/properties take precedence over the mixins but not sure when there are multiple mixins.  I guess this is one of main reasons that mixins are considered bad.  I think this way, even if the global mixin wasn't around then this will still work.

edwh:
I agree that it's clearer if you have a watch over a property that is 
defined in the component.  It's not simple to find out whether it is 
supposed to work, and therefore the code is more legible if we don't do 
that.

But I don't like the idea of doing that and then having a watch but not 
knowing which property it's watching on.

If the watch is working on the mixin property, then you don't need the 
local copy.

If the watch is working on the local copy, then clarity is improved by 
renaming it.

So I would favour renaming it and using a local copy.  Then nobody 
looking at the code in future is left confused over what's actually 
happening.

JayyajGH:
Yeah that makes sense. I agree. The watch does appear to work on the global mixin property but I don't like the idea of a component watching something in another component.  It's not very transparent.  I'll rename it.

JayyajGH:
Tweaked and added some comments

edwh:
This looks ok to me now. Undraft?

JayyajGH:
I've undrafted but I need to sort out the merges next...

JayyajGH:
Merge now complete so this is good to go

PR #339, opened 2020-11-27
Theming bootstrap

JayyajGH:
Demo of theming bootstrap


JayyajGH:
This now only deals with the primary and secondary buttons.

This will affect anything that uses the primary or secondary theme colours.  One of these I've noticed was the anchors which turned green to match the primary colour.  I've set up a default for this now to be a blue colour.    The slight worry is that there is something tucked away that uses one of these colours.  I've clicked on most pages and it generally looks ok.

PR #340, opened 2020-12-01
Feature/add accessible regions

JayyajGH:
Some more accessibility stuff.  Add header, main and aside regions.

### DIFF
--- a/components/BotLeftBox.vue
+++ b/components/BotLeftBox.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <nav>
     <b-card class="bg-transparent border-0">
       <div class="flex-wrap d-flex flex-row justify-content-between text-muted small list-unstyled">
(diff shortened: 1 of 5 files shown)


JayyajGH:
The only oddity here is how to mark up the pseudo-footer. i.e. the one with the links and social media links on the sidebar.  It's not really an aside and it's not really a footer.  It does contain a lot of useful links that aren't really available elsewhere so I think it qualifies as a nav element.

edwh:
Is this merged?  There are a bunch of changes to default.vue.

JayyajGH:
It was up to date but probably isn't now after the nchan stuff has gone to master.  Although github thinks it is...  I'll check

JayyajGH:
Well git is obviously cleverer than I thought as it was completely happy sorting out that merge by itself.

JayyajGH:
So this is now good to go...

PR #344, opened 2020-12-02
Refactor the main header out into a separate component

JayyajGH:
I've pulled all of the main header stuff out into a component.  It was easy.  Maybe too easy...  It's left the HTML in default.vue pretty bare which is nice.  I've not tidied the header in any way at all.  Just transplanted it into a separate component.

I've given it a quick test but thought I'd draft PR it so I can get your thoughts before giving a more thorough test.

How do I get the "show about me" option back in notifications?  Mine appears to have gone.  I just wanted to test that it still worked.

### DIFF
--- a/components/MainHeader.vue
+++ b/components/MainHeader.vue
@@ -1,0 +1,405 @@
+<template>
+  <header>
+    <!-- Navbar for large screens-->
+    <b-navbar id="navbar_large" toggleable="xl" type="dark" class="ourBack d-none d-xl-flex pl-1" fixed="top">
+      <b-navbar-brand to="/" class="p-0">
+        <b-img
+          class="logo mr-2"
+          height="58"
+          width="58"
+          rounded
+          :src="logo"
+          alt="Home"
+        />
+      </b-navbar-brand>
+      <client-only>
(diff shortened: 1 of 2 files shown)


JayyajGH:
Been giving this a more thorough test and it is looking ok.  I've also tested on iPhone 8(13), iPhone 11 (14) and Pixel 4.

Unrelated issue:
I've just noticed an issue with the notification panel being slightly off screen on a iPhone 8 (13) which is also there in live.  We've had this before.  I'm guessing that needs to be fixed?

JayyajGH:
Sorry I forgot to check in default.vue.  Done now

edwh:
Yes to fixing notification issue if you've noticed it.

JayyajGH:
I'll fix that independently of this I think as I need to look into in more fully as it's the default bootstrap functionality I think.

edwh:
Ok.

edwh:
My God, I think you've done it.  The layout is now down to a sane level of complexity.  I'll take this.

JayyajGH:
It kind of just fell out into the component.  Didn't have any troubles at all.  I'm still shocked!

PR #347, opened 2020-12-10
Feature/tidy index grid

JayyajGH:
Tidy up the grid layout


JayyajGH:
Various changes to tidy up the grid layout. 

The main point to discuss is I've currently removed the map from the medium screens as it was so small.

JayyajGH:
Another refactor following on from this would be to pull out the text stuff into a separate scss file so we can consolidate and tidy our use of fonts.  This is where I start talking about modular scales... :-)

JayyajGH:
At the moment a lot of the text and images on the map doesn't actually fit on the smaller screens.  I guess because it's too small to be able to centre and fit all of the bubbles on too.    I found about 300px high to be about the minimum required to see what is going on.

edwh:
Not looked at this yet.  But suppose we lost some of the text, would 
that help?  E.g. we could lose:

Got stuff you don't need? Looking for something?

Don't throw it away, give it away!

And we could shrink this:

Just looking?
Enter your location andSee what's being freegled near you.
...lose "just looking" and increase font size of "See.."

JayyajGH:
Yeah if you definitely want the map displayed in the middle screen sizes then we can make it fit.  I'll fit it in as is initially and we can go from there.

edwh:
Good, thank you.

JayyajGH:
See what you think of that now.  For now I've just added the map back in for the medium screens.  It's still not on the smaller ones and I haven't changed/removed any text.  Should I go further?

edwh:
I can't get that to have anything that overlaps, so it's definitely 
better.

I can get some resolutions (e.g. 1076x913) where the map is very narrow 
and there's a lot of wasted space on the right hand side.  That might be 
worth looking, somehow.

Yes, please go further.  I am worried that on mobile/tablet our landing 
page is basically just text, which doesn't get across the fact that 
there is genuine activity.

JayyajGH:
ok I'll make some more tweaks. Probably adding the map into one breakpoint below but not the smallest screens.  We can always give it a bit more height to make it usable. And then see about removing/moving some of the text.

JayyajGH:
The map now shows on all but the smallest of screens.  So basically anything that isn't a mobile.

JayyajGH:
I could make the map and the text 50/50 on the larger screens if you like.  It looks quite nice but obviously then the text has less focus.

edwh:
Screenshot?

JayyajGH:
image

edwh:
Doesn't look quite right - the map does make the text feel a bit 
squished.  Maybe 40:60 or something.

JayyajGH:
image

JayyajGH:
That is 40/60.  The map isn't loading at the moment but that should give an impression

edwh:
This is definitely better than what we have.  One more very minor tweak please - at some points the Mythic Beasts logo is very close to the charity footer.

image

That footer could be a little smaller, probably, but if it had a bit of padding at the top then it would look nicer.

JayyajGH:
I've just added some more spacing under the mythic beasts logo.  

What do you mean by "that footer could be a little smaller"?  The text is 16px at the moment.  That makes it great in terms of accessibility though.

edwh:
It's the small print about us being a charity, which means it's probably 
bigger than it needs to be.  People don't really need to read that, it 
just has to be there if they look for it.

JayyajGH:
The small print is now... small.

image

edwh:
Grand.

PR #348, opened 2020-12-10
Pull out the typography definitions

JayyajGH:
Pull out the typography definitions into a separate file.  This is phase one.

Phase two will be to sweep up all the ones scattered around in separate files and bring them in here.
Phase three will be to try and consolidate them.

This should lead to a more defined brand style that is common across the whole site.


JayyajGH:
ok cool.  I'll sort on Monday.

JayyajGH:
Merge done so should be good to go...

PR #350, opened 2020-12-15
First pass at consolidating the typography

JayyajGH:
This PR is grabbing out all of the scattered typography definitions and putting them in a single location.

I haven't made any attempt at consolidating or tidying them at the moment but at least they are now all in a single location.  I haven't touched anything to do with icons or buttons etc. as they are dealt with separately.  I've also not touched any of the definitions passed into bootstrap components at the moment.

The next stage will be tidying them up and creating something generic.  e.g. Only three sizes of body text, maybe something responsive, maybe use a modular scale for sizing, etc.

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -109,10 +109,6 @@ body.modtools {
   white-space: pre-line !important;
 }
 
-.fa-bh {
-  font-size: 100px;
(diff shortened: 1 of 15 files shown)


JayyajGH:
Actually

PR #351, opened 2020-12-15
Feature/typography tidy

JayyajGH:
This PR is grabbing out all of the scattered typography definitions and putting them in a single location.  Because it touches a lot of files I thought I would push this through to save creating a merge hell later on.

I haven't made any attempt at consolidating or tidying them at the moment but at least they are now all in a single location. I haven't touched anything to do with icons or buttons etc. as they will be dealt with separately. I've also not touched any of the definitions passed into bootstrap components at the moment.

The next stage will be tidying them up and creating something generic. e.g. Only three sizes of body text, maybe something responsive, maybe use a modular scale for sizing, etc.


edwh:
Bit worried about unintended consequences here.  This moves (for example) the name class from being scoped in quite an obscure component into a global style that would affect anyone who happened to use that class (including 3rd party components?).  Is it safe?

JayyajGH:
The name class would disappear as part of phase 2.  But I take your point.  All the other classes have pretty unique names but that one has the potential for clashing with something else when not scoped.  I'll sort that one out.

JayyajGH:
I've removed the "name" class and replaced it with a generic one.  There are a couple of others that have quite generic names too.  "head", "largest" and "medium".  I might do something about them before you merge too.

JayyajGH:
@edwh - I have now renamed any typography class that potentially could have a name clash with something else.

Everything is still weirdly named but I'll move on to creating a better generic type system next.

edwh:
Freegle likes weird, though, so maybe stick with yoghurt flavours, or 
whatever you've chosen.

JayyajGH:
Just merged all the other related PRs into this one

JayyajGH:
OK this is now phase one complete.  Moving all the obvious font definitions into a single file and consolidating them "a bit".  There is still quite a bit of work to do but this is at least a start.

edwh:
All looks good stuff.

fa-fw?

JayyajGH:
I see typography as font-family, font-size, font-weight and line-height.  fa-fw is the width and height.  There are quite a few little outsiders like this that I'm not sure about.  The font-awesome icons (not sure if this project does this as we're not using it directly) uses font-size to set the icon size.  To me this is outside of the general typography realm.  Ditto with the text in buttons etc.  That's just my personal opinion though.

The other issue is at the moment there are loads of <h1>, <h2> etc. that are just picking up the bootstrap font sizes as they don't use the new class names.  I would personally prefer all of them to be more explicit and always use a class.  Fundamentally I don't like the way that bootstrap directly associated the semantic header level with the font size.  Most of the popular normalise type CSS reset files will set all the header levels to 100% so you are forced to explicitly set them.

edwh:
Makes sense on icons.  It's possible that we will get some icons from 
Lawrence which are more likely to be images rather than fonts, and there 
will be other issues to do with icons, so we may need to have a separate 
iconography CSS section.

You need to be a bit careful about making everything explicit though.  
If we have a convention that every HTML element of type X needs a class 
from family Y then a) too much typing and b) we'll forget anyway.  It 
increases the barrier to entry, because it means you have to think about 
loads of things at the point when everything is new.  It's not 
unreasonable to expect standard HTML constructs to do sensible things, 
and you don't get much more standard than the header levels.

There's arguably a bug that headings combine the semantics with the 
visual, but that's a bug in HTML not considering accessibility enough in 
the first place, IMO.  You can't really fix that without removing native 
HTML elements entirely and moving to a purely component based structure, 
but I can't see that Bootstrap can do that because it doesn't mandate a 
front-end framework.

JayyajGH:
`<h1>` has no font size in itself.  It's purely semantic so it's not really a bug in HTML.  HTML doesn't define font sizes for headings and it shouldn't have any say so in it.  Browsers will set defaults but that's why people tend to use CSS resets and normalize files to remove it all.  Bootstrap uses a normalize file but then sets up it's own sizes so really the bug here is within browsers and bootstrap by linking semantics and design.  Then again they are both just trying to provide a basic set of styles for people to get something working quickly with minimal effort.

Isn't it the same with buttons though?  You will never just use a `<b-btn>` without adding a variant.  The developer will always need to know to add a variant.

The other option is to reset all of the header font sizes back to their normalize values e.g. 100% so probably 16px.  That way if a new developer adds a header without a styling class then it will be pretty obvious that it needs something done to it.  That way you are forced to remove the link between header semantics and styling.

I do realise that I'm coming from a bit of a purists point of view here though!

PR #353, opened 2020-12-16
Make the jobs more semantic - put them in a list

JayyajGH:
This is part of the overall work to tidy up the headers typography.  I'm currently trying to consolidate them down as we currently have about 6 different sizes with very little difference between some of them.

The jobs shouldn't really be headers.  They are just anchors in a list.  So I've put them in a list.


JayyajGH:
Closing this PR as I have now merged it into the other one

PR #356, opened 2021-01-01
Only show notifications when logged in

JayyajGH:
Hide the notifications icon on smaller screens when not logged in.


edwh:
I've seen larger changes.

JayyajGH:
Took ages. 😂 Although I did have to make the decision on whether to put the if condition on the parent block or each item. 🤔

edwh:
Probably best have a bit of a sit down and rest, then.

PR #357, opened 2021-01-09
Feature/infinite load

JayyajGH:
PR for discussion.  This is in no way fit for purpose for release.  It's just to prompt a discussion on the style as per our slack conversation.  Currently I have still left commented out code and unused code in here.

The key points are:

1. The observer component is the place where the actual intersection observation takes place
2. The JobsSideBar component is the place which uses the observer component to give the jobs an infinite load.

Ignore the InfiniteLoad component as this was just my place to check it was working.

### DIFF
--- a/components/JobsSidebar.vue
+++ b/components/JobsSidebar.vue
@@ -26,7 +26,9 @@
             <Job :summary="true" :job="job" />
           </li>
         </ul>
+        <Observer @intersect="intersected" />
         <client-only>
+          <!--
           <infinite-loading key="infinitejobs" @infinite="loadMore">
             <span slot="no-results">
               <notice-message v-if="!jobs || !jobs.length">
@@ -36,6 +38,7 @@
             <span slot="no-more" />
             <span slot="spinner" />
           </infinite-loading>
+        -->
         </client-only>
       </b-card-body>
     </b-card>
@@ -43,7 +46,9 @@
 </template>
 
 <script>
-import InfiniteLoading from 'vue-infinite-loading'
+// import InfiniteLoading from 'vue-infinite-loading'
+// import InfiniteLoad from '../components/InfiniteLoad'
+import Observer from './Observer'
 import Job from './Job'
 const NoticeMessage = () => import('~/components/NoticeMessage')
 const DonationButton = () => import('~/components/DonationButton')
@@ -52,7 +57,9 @@ export default {
   components: {
     Job,
     NoticeMessage,
-    InfiniteLoading,
+    Observer,
+    //  InfiniteLoading,
+    // InfiniteLoad,
     DonationButton
   },
   data: function() {
@@ -107,6 +114,13 @@ export default {
         $state.loaded()
       } else {
         $state.complete()
+      }
+    },
+    intersected() {
+      console.log('Getting more data...')
+
+      if (this.show < this.jobs.length) {
+        this.show += 3
       }
     }
   }
--- a/components/Observer.vue
+++ b/components/Observer.vue
@@ -15,12 +15,14 @@ export default {
   }),
   mounted() {
     const options = this.options || {}
+    // Use array destructing on the entry argument i.e. entries[0]
     this.observer = new IntersectionObserver(([entry]) => {
       if (entry && entry.isIntersecting) {
         this.$emit('intersect')
       }
     }, options)
 
+    // this.$el is the root element of the component
     this.observer.observe(this.$el)
   },
   destroyed() {
--- a/components/Observer.vue
+++ b/components/Observer.vue
@@ -1,0 +1,30 @@
+<template>
+  <div class="observer" />
+</template>
+
+<script>
+export default {
+  props: {
+    options: {
+      type: Object,
+      default: null
+    }
+  },
+  data: () => ({
+    observer: null
+  }),
+  mounted() {
+    const options = this.options || {}
+    this.observer = new IntersectionObserver(([entry]) => {
+      if (entry && entry.isIntersecting) {
+        this.$emit('intersect')
+      }
+    }, options)
+
+    this.observer.observe(this.$el)
+  },
+  destroyed() {
+    this.observer.disconnect()
+  }
+}
+</script>
--- a/components/InfiniteLoad.vue
+++ b/components/InfiniteLoad.vue
@@ -1,0 +1,49 @@
+<template>
+  <div>
+    <ul>
+      <li v-for="item in list" :key="item.id" class="list-item">
+        {{ item.title }}
+      </li>
+    </ul>
+    <Observer @intersect="intersected" />
+  </div>
+</template>
+
+<script>
+import Observer from './Observer'
+const a = require('axios')
+const axios = a.create({
+  timeout: 300000
+})
+
+export default {
+  components: {
+    Observer
+  },
+  data() {
+    return {
+      page: 1,
+      list: []
+    }
+  },
+  methods: {
+    async intersected() {
+      console.log('Getting more data...')
+
+      const ret = await axios.get(
+        '//hn.algolia.com/api/v1/search_by_date?tags=story',
+        {
+          params: {
+            page: this.page
+          }
+        }
+      )
+
+      console.log(this.list)
+
+      this.list = [...this.list, ...ret.data.hits]
+      this.page++
+    }
+  }
+}
+</script>


components/Observer.vue
```diff
@@ -0,0 +1,32 @@
...
+      type: Object,
+      default: null
+    }
+  },
+  data: () => ({
+    observer: null
+  }),
+  mounted() {
+    const options = this.options || {}
+    // Use array destructing on the entry argument i.e. entries[0]
```
edwh, line 18: Destructuring?

components/Observer.vue
```diff
@@ -0,0 +1,32 @@
...
+    this.observer = new IntersectionObserver(([entry]) => {
+      if (entry && entry.isIntersecting) {
+        this.$emit('intersect')
+      }
+    }, options)
+
+    // this.$el is the root element of the component
+    this.observer.observe(this.$el)
```
edwh, line 26: Do we know that this works if the element is created off-DOM initially and later added?

components/Observer.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <div class="observer" />
+</template>
+
+<script>
+export default {
+  props: {
```
edwh, line 7: The thing it is replacing has a distance parameter which is very useful, because it allows you to start fetching more content in time for it to be there when you need it.  Can we do that with this approach, e.g. by artificially positioning the element?

components/Observer.vue
```diff
@@ -0,0 +1,32 @@
...
+  },
+  data: () => ({
+    observer: null
+  }),
+  mounted() {
+    const options = this.options || {}
+    // Use array destructing on the entry argument i.e. entries[0]
+    this.observer = new IntersectionObserver(([entry]) => {
+      if (entry && entry.isIntersecting) {
+        this.$emit('intersect')
```
edwh, line 21: Might be better to name the component, events etc by the purpose not the implementation? Intersects is a bit opaque.

edwh: We could check at this point if the element is still in the DOM.  I worry that there might be a timing window between the observer firing and our callback, unless you know otherwise.  If it's not in the DOM, would it be better not to emit.

I had a method for that somewhere, probably another project, hops up the parent nodes to see if it ends at body though there will be better ways.

components/Observer.vue
edwh: More generally, I think I would want to preserve the existing interface.  It's a bad interface, but I wouldn't like to grapple both with subtle bugs in this, and bugs introduced by us changing the interface in loads of places.

components/JobsSidebar.vue
```diff
@@ -108,6 +115,13 @@ export default {
       } else {
         $state.complete()
       }
+    },
+    intersected() {
+      console.log('Getting more data...')
+
+      if (this.show < this.jobs.length) {
```
edwh, line 122: Here you're putting the burden of deciding whether to continue or stop on the parent component.  I quite like the original's approach of loaded() vs finished() as a way of moving that logic out of the parent.  

It could be even better, though:
* A clear way to wake the component up again later when things change.
* Some generic logic for handling the case where you've fetched some stuff and run out.  That needs some thought but we do it all over so it would be worth centralising.

Some of this might do in a higher level component than this one, ie this is a simple observer-wrapper and then there is infinite scroll logic which uses it.

edwh (commented):
Bunged in a few thoughts.  Some aren't new, but this is the place for the discussion now.

components/Observer.vue
JayyajGH: The intersection observer callback receives an array of entries.  We are only interested in the first one i.e entries[0].  We can easily get this out by destructuring the array to [entry].

This was something introduced in ES6:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

components/Observer.vue
JayyajGH: I can't think why it shouldn't but we would have to test it.  What exactly do you mean by "created off-DOM"?  Dynamically generated in JS and then appended to the DOM?

edwh:
You can create and render a Vue component and then attach it to the actual DOM later. We do that in the map marker code, I think.

edwh:
Yes, I just meant that I thought it was a typo - destructing should be destructuring

components/Observer.vue
JayyajGH: By "distance" do you mean that it starts fetching and loading when the bottom of the list is not yet displayed but is like 100px away?  There is a threshold property to the intersection observer but I think that only applies to whether it triggers at the top or bottom of the element being watched.  I'd need to investigate the best way to trigger when it's not yet quite displayed.
We can probably just create a similar component to the last one and wrap it all up in there probably using slots.  Just thinking off the top of my head but I'm sure we could make it work.

components/Observer.vue
JayyajGH: re. "Yes, I just meant that I thought it was a typo - destructing should be destructuring"

Doh.  It's hard to type with cold fingers. :-)

components/Observer.vue
JayyajGH: re. "You can create and render a Vue component and then attach it to the actual DOM later. We do that in the map marker code, I think."

The code runs on the mounted method and it's part of Vue so I'm pretty sure it would still work.  Easy enough to test though.

edwh:
Yes, that's what I mean. So in the infinite scroll case you can make sure you probably always have a screen full of stuff to scroll through, which means you never have to wait.

components/Observer.vue
JayyajGH: Yeah agreed.  This was the renamed version as I originally called it toInfinityAndBeyond.  Better than that but could be more semantic and less technical sounding.

So are you saying that potentially the intersection could happen, that triggers the data fetch and by the time that data comes back they could have scrolled back up the page?  In which case there would be no point adding it to the DOM?

components/JobsSidebar.vue
JayyajGH: I've just copied this logic from what it did before.  This was the previous method:

```
if (this.show < this.jobs.length) {
        this.show += 3
        $state.loaded()
      } else {
        $state.complete()
      }
```

There is definitely some improvements that could be made in this area.  I imagine that even when all the data is loaded that the observer will still run at the bottom of the list so we could detach it maybe.  I'm guessing that is what you mean by your second bullet point.

edwh:
I like that name.

Not quite:
- scroll down
- trigger api call
- scroll up
- scroll down
- trigger api call again
- first api call now completes
- other stuff starts happening 
- second api call now completes

Feels like there will be some cases where that would cause bugs.

components/Observer.vue
JayyajGH: From edwh:

I like that name.

Not quite:
- scroll down
- trigger api call
- scroll up
- scroll down
- trigger api call again
- first api call now completes
- other stuff starts happening
- second api call now completes

Feels like there will be some cases where that would cause bugs.

components/JobsSidebar.vue
edwh: No, I meant something else but it wasn't clear.  

Look at modMessagesPAge around line 139.  We get currentCount, then we fetch, then we check the currentCount against the (new) computed property from the store to decide whether we should call complete or loaded.   We have a bunch of logic like that which is a bit poor.

But on reflection that's part of having a better interface to the scroll, and perhaps a better interface from the server.  So ignore it for now.

components/Observer.vue
JayyajGH: The jobs section wouldn't have a problem as it just increments a variable so it doesn't matter which one completes first.

The other type will load data that it has fetched from an API.  I can't remember what the code does off the top of my head but I'm guessing it is some type of array push.  As JavaScript is single threaded (ignoring web workers) then it shouldn't matter which one returns first unless you are fetching the data in some kind of controlled order and displaying it as is rather than controlling the order in the view.  In C# you would probably use a mutex or something similar to lock the array whilst you were updating the data.  In JavaScript it should be safe.

Definitely worth adding to the list of tests though.  We can mock up some dummy functions that return at different times.  

This type of work would really benefit from having a unit test framework.  Nick was investigating that wasn't he?  Did he conclude anything?  I've always used Ava because it was the only one I could get working with Nuxt (and I think Nuxt recommend it) but I haven't looked for a while so there might now be others that work.

edwh: I can imagine more complex cases where it would matter.  

- Scroll and becomes visible, API call 1 starts
- Scroll and becomes visible, API call 2 starts
- API call 1 returns
- New "context" A set in client component for use on next API call (this is how we page through data)
- Scroll and becomes visible, API call 3 starts with new context A
- API call 3 completes with even newer context B
- New context B set in client component
- API call 2 completes with context A again
- Context A set in client component
- Next call will use context A rather than context B.

Will that matter?  I don't want to find out.  We just want to knock all of this kind of thing on the head by not triggering anything doing anything new if we are waiting for the callback to complete via finished() or loaded() became visible.

I agree that this is an excellent case for UT.  We didn't get as far with Nick as adopting something, so go with what you know.

JayyajGH: Crazy busy day today so I'll read this properly later.

I think it's worth having a look around to see which test frameworks play well with Nuxt now.  It's probably been over a year since I've looked at new ones.  Ava is alright. Seems to do the job but it always felt like I was fighting it when trying to add mocking libraries etc.

edwh:
At least if you're fire-fighting you'll stay warm.

edwh:
Closing as we've not continued on this and we will use [v3-infinite-loading ](https://github.com/oumoussa98/vue3-infinite-loading) in the rewrite.

PR #358, opened 2021-01-18
Feature/upgrade vue awesome

JayyajGH:
As discussed on slack

Try and encourage Vue to use at least 4.1.0 of vue-awesome and force an override of the vertical-align property.

### DIFF
--- a/assets/css/global.scss
+++ b/assets/css/global.scss
@@ -158,7 +158,7 @@ body.modtools {
 
   /* This is to resolve an issue with icon alignment. */
   /* It reverts a change made in 4.1.0 of vue-awesome */
-  vertical-align: middle;
+  vertical-align: middle !important;
 }
 
 .fa-fw {
--- a/package-lock.json
+++ b/package-lock.json
@@ -15476,18 +15476,11 @@
       "integrity": "sha512-WXjngEaNyNWFU9unUUdK5kGolCHgG3jmlUIgeRnKlHpskbgGjIE/HGTOWnMfLEqjYJl9DTzt/SKPWDoFVaND/A=="
     },
     "vue-awesome": {
-      "version": "4.0.2",
-      "resolved": "https://registry.npmjs.org/vue-awesome/-/vue-awesome-4.0.2.tgz",
-      "integrity": "sha512-TE9hVwyVrGnBetO/MJwZ/7qLO6Vgr/WrDtIA7vwNryoLdA2jfRcKRw29KacpU5xj3wzZhHfYJ2xJLxfirhmH5Q==",
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/vue-awesome/-/vue-awesome-4.1.0.tgz",
+      "integrity": "sha512-4n+hg8KIMrwjXV6sRHcRZd18Somih5j4Yk2ZOv95pnvDpzbBkIYW4ktfivhqgNt50m0zDjmeEWiy1iVLtcccfw==",
       "requires": {
-        "nanoid": "^2.1.6"
-      },
-      "dependencies": {
-        "nanoid": {
-          "version": "2.1.11",
-          "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-2.1.11.tgz",
-          "integrity": "sha512-s/snB+WGm6uwi0WjsZdaVcuf3KJXlfGl2LcxgwkEwJF0D/BWzVWAZW/XY4bFaiR7s0Jk3FPvlnepg1H1b1UwlA=="
-        }
+        "core-js": "^3.4.4"
       }
     },
     "vue-client-only": {
@@ -15496,7 +15489,7 @@
       "integrity": "sha512-arhk1wtWAfLsJyxGMoEYhoBowM87/i6HLSG2LH/03Yog6i2d9JEN1peMP0Ceis+/n9DxdenGYZZTxbPPJyHciA=="
     },
     "vue-draggable-resizable": {
-      "version": "github:Freegle/vue-draggable-resizable#4879847be38653d228dd78f811fc2326a53ed43d",
+      "version": "github:Freegle/vue-draggable-resizable#4c71a6c53eb6b2ee78f97a6a85757e14a1144c09",
       "from": "github:Freegle/vue-draggable-resizable"
     },
     "vue-eslint-parser": {
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "url": "^0.11.0",
     "vue": "^2.6.10",
     "vue-at": "^2.5.0-beta.2",
-    "vue-awesome": "^4.0.2",
+    "vue-awesome": "^4.1.0",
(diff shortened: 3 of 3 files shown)


JayyajGH:
@chriscant For info...

JayyajGH:
@chriscant - Did you ever get around to seeing if this worked in the app?  This change should have no effect on the website as that already appears to be picking up the latest version.

chriscant:
The current master code seems to use 4.0.2:
https://github.com/Freegle/iznik-nuxt/blob/master/package-lock.json

I haven't yet tested again in the app. Will give it a whirl.

JayyajGH:
Yeah I think the original issue was that because of the ^ in package.json, it appears that the web version was using 4.1.0 which effectively had a breaking change in it but the app was still using 4.0.2.  I was trying to make it more consistent.

JayyajGH:
The other option would be to lock it to 4.0.2 but then it stops us moving forward if there are any other fixes or updates.

chriscant:
package did have ^4.0.2 and the actually installed one as seen in package-lock was indeed 4.0.2. 

I upgraded to 4.1.0 and saw no problems. And added !important to global.css
vertical-align: middle !important;
and still also saw no problems.
I think the icons on buttons were and are aligned vertically OK.

Anyway, I think this one's good to go...

JayyajGH:
Originally it was you that raised this issue on Android but I can't actually see this problem happening anymore.  Can you still see it in live?

I'd still be tempted to use this change as it at least makes things specific as any new npm install is likely to install 4.1.0 due to our use of the caret.  This way it's definite.

chriscant:
Ed raised the issue which I did see in the app at the time. However it now seems to have gone. So I think we can safely upgrade to 4.1.0. We could check to see if the vert-align change is actually now needed - though I suspect that having it there is no problem.

JayyajGH:
I think the problem only appears when we are using 4.1.0.  If you are on 4.1.0 then you need the vertical-align: middle change due to this property being set in vue-awesome to a value that doesn't work for us.

Fundamentally we either need to:

1. Enforce using 4.0.2 by setting it explicitly in package.json.  The downside of this is we don't benefit from any future enhancements.
2. Enforce using (at least) 4.1.0 and use the override of vertical-align in global.scss

We've seen a difference in versions used on desktop and the app with the same package.json setting so I think doing nothing would be a bad idea. With the current settings then everywhere SHOULD be picking up the highest minor version which is 4.1.0. (unless I've misinterpreted how semver works in npm)

I think my preference would be for (2) as then we will at least be consistent.

@edwh - Which would you prefer?

edwh:
2 sounds better to me, because otherwise we are stuck on 4.0.2 forever, 
as you said earlier.

It's unclear whether the problem still applies but I don't mind having 
an override in global.scss just in case.

It's probably a bad idea in general for us to be using ^, right?

From: "Jason Harris" ***@***.***>
To: "Freegle/iznik-nuxt" ***@***.***>
Cc: "Edward Hibbert" ***@***.***>; "Mention" 
***@***.***>
Sent: 12/03/2021 13:35:16
Subject: Re: [Freegle/iznik-nuxt] Feature/upgrade vue awesome (#358)

>
>I think the problem only appears when we are using 4.1.0. If you are on 
>4.1.0 then you need the vertical-align: middle change due to this 
>property being set in vue-awesome to a value that doesn't work for us.
>
>Fundamentally we either need to:
>
>Enforce using 4.0.2 by setting it explicitly in package.json. The 
>downside of this is we don't benefit from any future enhancements.
>Enforce using (at least) 4.1.0 and use the override of vertical-align 
>in global.scss
>We've seen a difference in versions used on desktop and the app with 
>the same package.json setting so I think doing nothing would be a bad 
>idea. With the current settings then everywhere SHOULD be picking up 
>the highest minor version which is 4.1.0. (unless I've misinterpreted 
>how semver works in npm)
>
>I think my preference would be for (2) as then we will at least be 
>consistent.
>
***@***.*** <https://github.com/edwh> - Which would you prefer?

JayyajGH:
Generally ^ should be fine as it should only ever upgrade minor versions which really means they shouldn't contain breaking changes.  I guess the definition of breaking change is loose.  The alternative is manually upgrading things but we all know that that would never happen.  NPM must want you to use ^ as that is the default.

edwh:
Ok, stick with it.

I do upgrade things occasionally.  When we move to Nuxt 3...

From: "Jason Harris" ***@***.***>
To: "Freegle/iznik-nuxt" ***@***.***>
Cc: "Edward Hibbert" ***@***.***>; "Mention" 
***@***.***>
Sent: 12/03/2021 14:28:57
Subject: Re: [Freegle/iznik-nuxt] Feature/upgrade vue awesome (#358)

JayyajGH:
In that case this PR should be good to go!

JayyajGH:
@chriscant @edwh - I've pulled master and did an npm install and I now get 4.1.0 locally.  I think it's worth checking that this happens during the website and app build process to confirm we are using 4.1.0 so we know where we are with this.  It's easy to test once live as the new version of vue-awesome sets vertical-align to be -0.125em.

I've noticed that in live and in this "fixed" version some of the icons still aren't lined up properly.  This change doesn't really fix anything; it just makes things align in the same way as 4.0.2.

edwh:
I wish It Just Worked.

I'll check the version next time I install.

From: "Jason Harris" ***@***.***>
To: "Freegle/iznik-nuxt" ***@***.***>
Cc: "Edward Hibbert" ***@***.***>; "Mention" 
***@***.***>
Sent: 12/03/2021 14:48:50
Subject: Re: [Freegle/iznik-nuxt] Feature/upgrade vue awesome (#358)

>
***@***.*** ***@***.***

chriscant:
The icons look OK to me, but mibbie I'm not fussy. Can you share some screenshots on slack?
Alternatively: can you fix the icons that aren't aligned.
Ta, Chris

JayyajGH:
image

JayyajGH:
This is on desktop.  The "OFFER something" icon is fairly well aligned but the "Your OFFERs" icon is slightly too low.

All the icons are added in the same way so it's likely to be the context they are used which is causing the issue.  i.e. we would need to investigate that.  I'll take a quick look on Monday...

edwh:
I've seen two things with icons:
Vertical alignment.  I think I've fixed some of those, probably in 
different places in different places (E.g. flex-column + 
justify-content-center).  But it feels like they should just work, since 
they're font based.
Width.  If you have icons one below the other they can look stupid, 
because the width varies.  I've used fa-fw in some places but again it 
feels like there should be a consistent solution.
Along similar lines, it's a pain that when you have some text and then a 
button with variant="link", the alignment is off.

From: "Jason Harris" ***@***.***>
To: "Freegle/iznik-nuxt" ***@***.***>
Cc: "Edward Hibbert" ***@***.***>; "Mention" 
***@***.***>
Sent: 12/03/2021 18:06:15
Subject: Re: [Freegle/iznik-nuxt] Feature/upgrade vue awesome (#358)

PR #360, opened 2021-02-08
Feature/seo page tweaks

JayyajGH:
How about this?  I've tried to keep it subtle...

### DIFF
--- a/components/RandomImage.vue
+++ b/components/RandomImage.vue
@@ -5,7 +5,7 @@
         v-if="image"
         :src="image"
         thumbnail
-        class="image"
+        class="image fade-in"
         generator-unable-to-provide-required-alt=""
         @error.native="brokenImage"
       />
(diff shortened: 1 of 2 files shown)


JayyajGH:
The row of ten images looks a bit odd on a mobile too.  Do they need to be shown at all?  Or maybe reduce the number of them.  They are very small at the moment.

edwh:
Yes, they probably look a bit odd on mobile.  I want to get across the 
impression that there's real stuff, and also put some different content 
on the pages to try to fool Google into not noticing they're the same.  
What would be nicer would be to show a different number based on screen 
size.

edwh:
The fade looks good, though; shall I take this as is?

JayyajGH:
Yeah take this if you're happy with it.  It's not just fade though.  It's fade and slight scale.  These things are important. :-)

JayyajGH:
I'll look at tweaking the mobile view tomorrow if I get 5 mins spare!

edwh:
I did notice :-)

PR #361, opened 2021-02-08
Add robots file

JayyajGH:
An initial robots file

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,0 +1,5 @@
+User-agent: *
+Disallow:
+
+# Sitemap files
+Sitemap: https://ilovefreegle.org/sitemap.xml


edwh:
Assuming you've done the right thing, then that looks correct :-).

JayyajGH:
That's the most basic of robots file that I've used elsewhere.  It basically allows any bot to crawl it.  The main bit is the link to the xml sitemap

JayyajGH:
You can point a tester tool at it once it's live so give me a shout when it is and I'll double check it's ok

JayyajGH:
It is probably worth looking at the website folder to see what is accessible. I've not familiar with exactly what is deployed so probably best for you to do this bit.  I've set "disallow" to be empty which means everything is visible.  You would usually use this section for stopping bots looking at wp-admin folders and similar.

edwh:
Live.  I'll tweak the URL to add www. soon.

JayyajGH:
Ah dang.  I double checked the live url but forgot that chrome removes the www subdomain from it!

JayyajGH:
Just about to create a PR to add the subdomain

edwh:
No need, I did that.

PR #363, opened 2021-02-10
Uncapitalise the text as it is read out strangely

JayyajGH:
When written in caps a sr reads out OFFER and WANTED as the individual letters.

### DIFF
--- a/components/AdaptiveMap.vue
+++ b/components/AdaptiveMap.vue
@@ -128,7 +128,7 @@
         <JobsTopBar v-if="jobs" />
 
         <h2 class="sr-only">
-          List of WANTEDs and OFFERs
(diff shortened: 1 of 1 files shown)


edwh:
I don't want to do this - we want them as OFFER and WANTED for existing users.  So does that mean we need to use sr-only etc?

JayyajGH:
This is just screen reader only so won't be displayed anywhere

JayyajGH:
<h2 class="sr-only">
          List of wanteds and offers
        </h2>

edwh:
Doh.

JayyajGH:
This is part of the testing I'm doing on the new message stuff.

PR #364, opened 2021-02-28
Fix UI inconsistency

tamarameghla:


edwh:
Hello Tamara - you're new!

You're right, there's inconsistency in the sidebars, and some confusion about what is in the layout vs what is in each page.

I'm not quite sure if you're done yet?  Let me know.

components/ProfileHeader.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <b-media>
+  <b-media v-if="user">
```
edwh, line 2: This v-if is deliberate.  What's the justification for removing it?

pages/help.vue
```diff
@@ -1,7 +1,10 @@
 <template>
-  <div class="bg-white">
+  <div class="">
```
edwh, line 2: The Help page is for people who are having trouble.  We therefore want to keep it simple:

- white background for readability
- no sidebars to allow people to focus on the actual help

This means there is inconsistency with other pages, but deliberately so.

pages/promote.vue
```diff
@@ -1,6 +1,8 @@
 <template>
-  <b-row class="m-0 bg-white">
-    <b-col cols="0" md="3" />
+  <b-row class="m-0">
```
edwh, line 2: ...but it's ok to be more consistent on this page.

edwh (changes requested):

edwh:
I'll close this PR soon, but if you're still active, please have a look at the feedback above.

PR #385, opened 2021-11-17
Fix typo

ngm:
Typo: "It's is public"

It could change to either "It is" or "It's".

But "It's" is used frequently elsewhere in the copy, so I went with that.


edwh:
Fancy seeing you here...

PR #399, opened 2022-05-01
Fix case consistency for rating (thumbs) text

JohnVeness:
This fixes the case inconsistency between "X freeglers gave them a thumbs up" and "Y freeglers gave them a thumbs Down".


edwh:
Thanks.

PR #400, opened 2022-05-01
Fix hover on the last seen date

JohnVeness:
This fixes the text that is shown if you hover over the last seen time of a member.

Please note that I haven't actually tested this! I am assuming that removing the quotes will show the actual timestamp on hover. If not, then maybe just remove the title attribute completely.


edwh:
Thanks.  The fix isn't quite right - in Vue you need to put a : before the quotes to force it to evaluate what is inside them.

The client dev environment is pretty easy to get going - see the README.  Fancy trying it?

